PR #1, opened 2026-02-08
Add organization billing model and members/billing console flows

RhysSullivan:
## Summary
- add an org-first Convex data model and APIs for organizations, memberships, invites, workspace ownership, and seat-based Stripe billing sync
- integrate `@convex-dev/stripe` into Convex config and wire webhook registration, plus billing actions/queries for checkout, portal, and seat sync retries
- add members and billing routes/views in the web app and update session/navigation flows to support org-aware workspace selection and invite/billing provider state
- move task/tool frontend calls to Convex executor functions (`executor` / `executorNode`) and remove the old web `executor-client` wrapper

## Validation
- ran `bun run convex:codegen` in `executor`
- ran `bun test` in `executor`

PR #2, opened 2026-02-08
Add organization billing model and members/billing console flows

RhysSullivan:
Supersedes #1 with correct base after main push.


executor/convex/app.ts
```diff
@@ -0,0 +1,18 @@
+import { query } from "./_generated/server";
+
+const workosEnabled = Boolean(process.env.WORKOS_CLIENT_ID && process.env.WORKOS_API_KEY);
+
+export const getClientConfig = query({
```
RhysSullivan, line 6: probably bad

executor/convex/auth.ts
```diff
@@ -54,6 +54,151 @@ async function getWorkspaceByWorkosOrgId(ctx: { db: any }, workosOrgId: string)
...
+  const baseSlug = slugify(baseName);
+  const existing = await ctx.db
+    .query("organizations")
+    .withIndex("by_slug", (q: any) => q.eq("slug", baseSlug))
+    .unique();
+  if (!existing) {
+    return baseSlug;
+  }
+
+  for (let i = 0; i < 20; i += 1) {
```
RhysSullivan: make reusable

executor/convex/auth.ts
```diff
@@ -54,6 +54,151 @@ async function getWorkspaceByWorkosOrgId(ctx: { db: any }, workosOrgId: string)
...
+    if (!collision) {
+      return candidate;
+    }
+  }
+
+  return `${baseSlug}-${Date.now()}`;
+}
+
+async function upsertOrganizationMembership(
+  ctx: { db: any },
```
RhysSullivan: dont use any

executor/convex/auth.ts
```diff
@@ -54,6 +54,151 @@ async function getWorkspaceByWorkosOrgId(ctx: { db: any }, workosOrgId: string)
...
+    invitedByAccountId: args.invitedByAccountId,
+    joinedAt: args.status === "active" ? args.now : undefined,
+    createdAt: args.now,
+    updatedAt: args.now,
+  });
+}
+
+async function ensureOrganizationForWorkspace(
+  ctx: { db: any },
+  args: {
```
RhysSullivan: can you import thistype from somewhere

executor/convex/auth.ts
```diff
@@ -54,6 +54,151 @@ async function getWorkspaceByWorkosOrgId(ctx: { db: any }, workosOrgId: string)
...
+  for (const invite of pendingInvites) {
+    await ctx.db.patch(invite._id, {
+      status: "accepted",
+      acceptedAt: args.acceptedAt,
+      updatedAt: args.acceptedAt,
+    });
+  }
+}
+
 async function ensurePersonalWorkspace(
```
RhysSullivan, line 151: lets solve this by just making organizationId not nullable in our schema instead then its always required to be defined  and we can get rid of all this ensure crap in the codebase

PR #18, opened 2026-03-11
Add API key support to the control plane client

bam4564:
## Summary
- add optional apiKey support to createControlPlaneClient
- send both the bearer token and account header when present
- keep the client helper usable for authenticated local control-plane calls

## Why
The control-plane API expects bearer auth for local workspace operations, but the exported client helper only attached the account header.


bam4564:
Closing this. Codex got confused lol

PR #116, opened 2026-04-08
add promise-based @executor/sdk and plugin subpath exports

RhysSullivan:
Create packages/published/sdk with a promise-based wrapper around
@executor/core for external consumers. Add public.ts subpath exports
to all plugin packages (. for promise SDK, ./core for Effect internals).
Configure tsup builds and publishConfig for npm publishing.


RhysSullivan:
Superseded by #142 — consolidated @executor-js/core + @executor-js/sdk into one package with dual entry points.

PR #135, opened 2026-04-09
feat(mcp): support manual headers for remote sources

aryasaatvik:
## Summary
- extract reusable secret-backed header auth UI from the OpenAPI add flow
- reuse the shared header auth UI in OpenAPI
- add remote MCP auth mode selection with secret-backed auth headers
- add plaintext additional headers for remote MCP sources

## Verification
- bun --cwd packages/react typecheck
- bun --cwd packages/plugins/openapi typecheck
- bun --cwd packages/plugins/mcp typecheck


RhysSullivan:
rebased & made graphql used the same input

PR #138, opened 2026-04-09
fix: publish @executor-js library packages on release

RhysSullivan:
The `@executor-js/core`, `@executor-js/sdk`, and `@executor-js/plugin-*`
packages were never reachable by the release workflow:

- All seven `@executor-js/plugin-*` packages were still in the changeset
  `ignore` list, so `changeset version` never bumped them.
- The root `build:packages` script still had a stale `@executor/plugin-*`
  glob, so plugin dist output was never built for publish.
- `changesets/action@v1` was invoked with `version:` only — no `publish:`
  step — and the CLI publish workflow only ships the executor wrapper.

Fix by removing the public plugins from the ignore list, correcting the
build glob, adding `scripts/publish-packages.ts` (which packs + publishes
every public @executor-js package whose version is not already on npm),
and wiring a new publish step into `release.yml` that runs whenever the
Version Packages PR is merged. Library publish now runs in parallel with
the existing CLI publish dispatch.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -19,13 +19,6 @@
     "@executor/ir",
     "@executor/runtime-deno-subprocess",
     "@executor/runtime-quickjs",
-    "@executor-js/plugin-file-secrets",
-    "@executor-js/plugin-google-discovery",
-    "@executor-js/plugin-graphql",
-    "@executor-js/plugin-keychain",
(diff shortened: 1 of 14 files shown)


RhysSullivan:
Superseded by #142 — consolidated @executor-js/core + @executor-js/sdk into one package with dual entry points.

PR #158, opened 2026-04-10
fix: two independent hanging-write bugs in the execution engine and local Bun server

aryasaatvik:
## Summary

Fixes two independent causes of hung non-SAFE tool calls that could still complete upstream, producing phantom writes where the mutation succeeded but the caller never received a response.

## Changes

- Use `Effect.forkDaemon` for pausable sandbox executions.
  - `executeWithPause` and `resume` are often driven by separate top-level `Effect.runPromise` calls, such as HTTP requests or CLI invocations.
  - A regular `Effect.fork` ties the sandbox fiber to the first caller scope, so it can be interrupted as soon as the paused result is returned.
  - `forkDaemon` lets the paused execution survive across the resume boundary.

- Set `Bun.serve({ idleTimeout: 0 })` for the local server.
  - Bun's default 10s idle timeout is too short for MCP elicitation and pause/resume flows involving human approval.
  - Disabling the Bun-level socket timeout prevents the server from closing the connection mid-flow.
  - MCP and HTTP clients still enforce their own request timeouts.

## Tests

- Added a regression test for a single-elicit tool across separate `runPromise` boundaries.
- Verified the regression fails with `Effect.fork` and passes with `Effect.forkDaemon`.

## Validation

- [x] `bun run typecheck` and `bun x vitest run` in `packages/core/execution`
- [x] `bun run typecheck` in `apps/local`
- [x] End-to-end check against Gmail `labels.create`

### DIFF
--- a/apps/local/src/serve.ts
+++ b/apps/local/src/serve.ts
@@ -106,10 +106,8 @@ export async function startServer(opts: StartServerOptions = {}): Promise<Server
   const server = Bun.serve({
     port,
     hostname: "127.0.0.1",
-    // Disable Bun's default 10s idle timeout. `/mcp` elicitation round-trips
-    // and `/api/executions` pause/resume can legitimately keep the socket
-    // idle longer than that. `0` is the "no timeout" sentinel per Bun's
-    // uSockets; MCP/HTTP clients enforce their own per-request timeouts.
+    // Disable Bun's default 10s idle timeout. MCP elicitation and pause/resume
+    // can idle longer during human approval; `0` disables the socket timeout.
     idleTimeout: 0,
     routes: { ...staticRoutes },
     async fetch(req) {
--- a/apps/local/src/serve.ts
+++ b/apps/local/src/serve.ts
@@ -106,6 +106,17 @@ export async function startServer(opts: StartServerOptions = {}): Promise<Server
   const server = Bun.serve({
     port,
     hostname: "127.0.0.1",
+    // Bun's default idleTimeout is 10 seconds (see `ServerConfig.zig`), which
+    // is too short for the MCP streamable-HTTP transport on `/mcp` and the
+    // executor's own `/api/executions` pause/resume flow: a single tool call
+    // may legitimately spend longer than that awaiting an elicitation
+    // response, running a long sandboxed computation, or holding the
+    // connection open while a human approves. `idleTimeout: 0` disables the
+    // idle timeout entirely — verified in `bun-usockets/src/socket.c`'s
+    // `us_socket_timeout`, where `seconds == 0` sets the internal timeout
+    // field to the sentinel 255 which never fires. The MCP/HTTP clients are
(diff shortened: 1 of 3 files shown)


packages/core/execution/src/engine.ts
```diff
@@ -353,7 +357,7 @@ export const createExecutionEngine = (config: ExecutionEngineConfig): ExecutionE
         });
 
       const invoker = makeFullInvoker(executor, { onElicitation: elicitationHandler });
-      fiber = yield* Effect.fork(codeExecutor.execute(code, invoker));
+      fiber = yield* Effect.forkDaemon(codeExecutor.execute(code, invoker));
```
aryasaatvik, line 357: This is the core lifetime fix. A paused execution can cross multiple host calls: `/api/executions` returns the pause, then `/api/executions/:id/resume` joins the same sandbox later. With `Effect.fork`, the sandbox remains a child of the first `runPromise`, so returning the paused response can interrupt it before the caller has a chance to resume. `forkDaemon` moves that sandbox fiber into the global scope, which matches the execution ID lifecycle instead of the request lifecycle.

packages/core/execution/src/tool-invoker.test.ts
```diff
@@ -346,4 +362,58 @@ describe("pause/resume with multiple elicitations", () => {
       }),
     { timeout: 10000 },
   );
+
+  // Regression: each engine call must run in its own top-level
+  // `runPromise` to reproduce the HTTP shape. Uses a single-elicit tool so
+  // nothing can accidentally unstick a dead-fiber race via a second
+  // elicitation (which is what masks the bug in the `multiApproval` test
+  // above).
+  it(
```
aryasaatvik, line 369: The important detail here is using a normal async Vitest test instead of `it.effect`. That makes `executeWithPause` and `resume` run through separate top-level `Effect.runPromise` calls, matching HTTP/CLI usage. The single-elicit tool avoids the old multi-elicit test accidentally unblocking the dead-fiber race with a second pause signal.

apps/local/src/serve.ts
```diff
@@ -106,6 +106,11 @@ export async function startServer(opts: StartServerOptions = {}): Promise<Server
   const server = Bun.serve({
     port,
     hostname: "127.0.0.1",
+    // Disable Bun's default 10s idle timeout. `/mcp` elicitation round-trips
+    // and `/api/executions` pause/resume can legitimately keep the socket
+    // idle longer than that. `0` is the "no timeout" sentinel per Bun's
+    // uSockets; MCP/HTTP clients enforce their own per-request timeouts.
+    idleTimeout: 0,
```
aryasaatvik, line 111: This covers the other hang path. Bun defaults HTTP connections to a 10s idle timeout, which is short enough to cut off MCP elicitation while a human approval UI is still round-tripping. Setting this to `0` disables the server-level socket cutoff; MCP and callers still keep their own logical request timeouts.

PR #164, opened 2026-04-10
feat(react/mcp-install-card): show agent provider logos

mrzmyr:
## Summary

- Surfaces Cursor, Claude, and OpenCode logos on the Connect-to-Agent install card so users immediately understand which clients they can paste the command into.
- Adds `@lobehub/icons` as a dependency of `@executor/react` for the brand glyphs.

## Changes

- `packages/react/src/components/mcp-install-card.tsx`
- `packages/react/package.json`
- `bun.lock`

## Notes

`bun.lock` has a large diff because `@lobehub/icons` pulls in several transitive deps (ant-design, emoji-mart, dnd-kit, emotion). All additions are downstream of the single new dependency.


mrzmyr:
@RhysSullivan ready to merge

PR #165, opened 2026-04-10
feat(react): add CardStack primitive and adopt it across surfaces

mrzmyr:
## Summary

Introduces a new `<CardStack>` primitive that standardizes the grouped-card layout used across the app — one container style, consistent padding, hover affordances, and slotted title / description / action regions.

Adopts it in four places that were previously reinventing the same pattern with hand-rolled `<div className="rounded-lg border ...">` markup:

- **Secrets page** — each secret row becomes a `<CardStackEntry>` with title, description, and action slots.
- **Schema explorer** — wrapped in `<CardStack>` / `<CardStackHeader>` with a computed field count.
- **Tool detail** — schema and TypeScript sections use CardStack headers, plus a small `EmptySection` helper for consistent empty states.
- **Tool tree** — the recursive tree rendering is flattened into a row-based layout inside a CardStack wrapper.

## Changes

- `packages/react/src/components/card-stack.tsx` (new, primitive)
- `packages/react/src/pages/secrets.tsx`
- `packages/react/src/components/schema-explorer.tsx`
- `packages/react/src/components/tool-detail.tsx`
- `packages/react/src/components/tool-tree.tsx`


mrzmyr:
@RhysSullivan ready to merge

PR #167, opened 2026-04-10
feat(sources): expose upstream url and render SourceFavicon

mrzmyr:
## Summary

Adds an optional `url` field to the `Source` schema and wires each built-in source plugin to populate it. Introduces a new `<SourceFavicon>` component that derives a favicon from the source URL's base origin, and uses it in both app shells to give the sidebar a recognizable "title-card" feel.

## Changes

**Schema plumbing** — `url?: string` on `Source`
- `packages/core/sdk/src/sources.ts`
- `packages/core/api/src/sources/api.ts`
- `packages/core/api/src/handlers/sources.ts`

**Plugin population**
- `packages/plugins/mcp/src/sdk/plugin.ts` — remote transport endpoint
- `packages/plugins/graphql/src/sdk/plugin.ts` — configured endpoint
- `packages/plugins/openapi/src/sdk/plugin.ts` — spec `baseUrl`
- `packages/plugins/google-discovery/src/sdk/plugin.ts` — `rootUrl`

**UI**
- `packages/react/src/components/source-favicon.tsx` (new)
- `apps/cloud/src/web/shell.tsx` — sidebar favicon + "Sources" label
- `apps/local/src/web/shell.tsx` — sidebar favicon + "Sources" label

## Notes

`SourceFavicon` shows a skeleton while loading and falls back gracefully if the host doesn't serve a favicon, so sources without a URL (e.g. stdio MCP transports) still render cleanly.

PR #168, opened 2026-04-10
feat(react): add global command palette (⌘K)

mrzmyr:
## Summary

- Introduces `<CommandPalette>`, a shared keyboard-driven navigator that lists sources grouped by plugin.
- Wired into both the cloud and local app shells so users can jump between sources without reaching for the sidebar.
- Plugin-driven — each shell passes the set of source plugins it has loaded and the palette renders entries grouped by plugin kind.

## Changes

- `packages/react/src/components/command-palette.tsx` (new)
- `apps/cloud/src/web/shell.tsx`
- `apps/local/src/web/shell.tsx`

## Notes

The shell.tsx files are also touched by #167 (SourceFavicon). The hunks are disjoint, so whichever PR merges first will cleanly rebase the other.


mrzmyr:
@RhysSullivan ready to merge

PR #241, opened 2026-04-14
[codex] Improve dynamic worker error formatting

heyglassy:
## Summary

- prevent object-shaped dynamic worker errors from collapsing to `[object Object]`
- preserve normal `Error.message` behavior for standard errors
- add regression coverage for thrown objects and object-shaped tool failures in the Cloudflare dynamic worker tests

## Root Cause

The dynamic worker runtime formatted execution failures with `err.message ?? String(err)`. When sandboxed code or a tool failure threw a plain object, that fell through to `String(err)` and surfaced as `[object Object]`.

## Validation

- `cd packages/kernel/runtime-dynamic-worker && bun x vitest run src/invocation.test.ts --reporter=verbose`
- `cd packages/kernel/runtime-dynamic-worker && bun x vitest run src/executor.test.ts --reporter=verbose`

## Screenshot Context

image
CleanShot 2026-04-14 at 11 46 39@2x

PR #244, opened 2026-04-14
[codex] Refactor dynamic worker error transport with Effect causes

heyglassy:
## Summary

- replace the dynamic worker's string-only error transport with a structured internal envelope derived from Effect causes
- preserve the existing outward execution contract and user-facing error strings while keeping structured failures intact across the worker boundary
- rethrow structured tool failures through the worker transport so plain objects render as JSON instead of collapsing to `[object Object]`

## Root Cause

The dynamic worker runtime flattened failures at the Workers RPC boundary with ad hoc string conversion. That discarded Effect cause structure and collapsed object-shaped failures to `[object Object]` before the execution engine could format them.

## Validation

- `cd packages/kernel/runtime-dynamic-worker && bun x vitest run`

## Notes

- supersedes #241
- this is a clean refactor from `main`, not a follow-up on top of the minimal-fix branch


packages/kernel/runtime-dynamic-worker/src/executor.ts
```diff
@@ -54,20 +51,169 @@ export type DynamicWorkerExecutorOptions = {
...
+    value === null ||
+    typeof value === "string" ||
+    typeof value === "number" ||
+    typeof value === "boolean"
+  ) {
+    return value;
+  }
+
+  try {
+    return JSON.parse(JSON.stringify(value)) as SerializedWorkerErrorValue;
```
RhysSullivan, line 116: do this in effect, use effet schema etc is kind of a mess

heyglassy: Updated this to use `effect/Schema` for the normalized error object and RPC envelope in `executor.ts`, instead of the ad hoc response/type plumbing. The transport behavior stays the same, and `cd packages/kernel/runtime-dynamic-worker && bun x vitest run` still passes.

PR #253, opened 2026-04-15
fix: make tests OS-Agnostic

mynameistito:
A few tests were breaking on Windows during local dev. Here's what was wrong and what changed.

## What broke

**`xdg.test.ts`** - path assertions used hardcoded forward-slash strings like `"/home/user/.local/share"`. On Windows, `path.join` uses backslashes, so those comparisons always failed. Fixed by using `join()` in the assertions to match whatever the runtime produces. Same issue with the `C:\Users\user` assertion - replaced it with per-segment checks so it works regardless of separator.

**`runtime-secure-exec` tests** - `secure-exec-v8` doesn't ship a Windows binary, so the whole suite was exploding with `ENOENT`. Wrapped the describe block with `describe.skipIf(process.platform === "win32")` so it skips cleanly instead of crashing.

**`release-bootstrap-smoke.test.ts`** - `process.kill("SIGKILL")` throws on Windows ("Unknown signal: SIGKILL"). Replaced both kill calls with no-argument `kill()`, which terminates the process on every platform.

**`runtime-deno-subprocess` network test** - the test fetching `https://example.com` was returning null on Windows, probably a firewall thing. Marked it `skipIf(CI || win32)` since it's an external network call anyway and not something the test suite should depend on in CI.

## What didn't change

All test logic is untouched. No skips were added where the underlying code has real Windows support - the secure-exec skip is there because the binary flat-out doesn't exist for that platform.

### DIFF
--- a/packages/kernel/runtime-deno-subprocess/src/index.test.ts
+++ b/packages/kernel/runtime-deno-subprocess/src/index.test.ts
@@ -168,7 +168,8 @@ skipUnlessDeno("runtime-deno-subprocess", () => {
     }),
   );
 
-  it.effect("network access can be allowed via permissions", () =>
+  // Skipped in CI and on Windows — outbound HTTPS may be blocked by firewall/policy
+  it.effect.skipIf(process.env["CI"] === "true" || process.platform === "win32")("network access can be allowed via permissions", () =>
     Effect.gen(function* () {
       const executor = makeDenoSubprocessExecutor({
         permissions: {
--- a/packages/kernel/runtime-secure-exec/src/index.test.ts
+++ b/packages/kernel/runtime-secure-exec/src/index.test.ts
@@ -1,5 +1,4 @@
-import { expect, it } from "@effect/vitest";
-import { describe } from "vitest";
+import { describe, expect, it } from "@effect/vitest";
 import { assertInclude } from "@effect/vitest/utils";
 import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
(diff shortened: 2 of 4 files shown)


mrzmyr (commented):
## Claude Code Review (Opus 4.6)

Sensible changes overall — skips are well-justified and the xdg `join()` fix is correct.

One real issue:

### `kill("SIGKILL")` → `kill()` loses the force-kill escalation

In `release-bootstrap-smoke.test.ts`, the original pattern is intentional: try graceful `SIGTERM`, wait 5s, then force-kill with `SIGKILL`. Replacing the second call with bare `kill()` just sends `SIGTERM` again on Unix, which does nothing if the process already ignored the first one. You can end up with a zombie process hanging the test suite.

```ts
// before (correct escalation)
webProcess.kill("SIGTERM");  // ask nicely
// ...5s timeout...
webProcess.kill("SIGKILL");  // force

// after (sends SIGTERM twice — no escalation)
webProcess.kill();  // ask nicely
// ...5s timeout...
webProcess.kill();  // ask nicely again
```

A cross-platform fix would be:

```ts
if (webProcess.exitCode === null) {
  process.platform === "win32"
    ? webProcess.kill()           // Windows: always terminates
    : webProcess.kill("SIGKILL"); // Unix: force-kill
}
```

mynameistito:
thanks @mrzmyr fixed here https://github.com/RhysSullivan/executor/pull/253/changes/0ca0c73b7ebdb3e0743c45ec3605ff5c32fa0196

mynameistito:
ty for the merge @RhysSullivan

PR #259, opened 2026-04-15
[codex] Avoid duplicate built-in source badge

heyglassy:
## What changed
- removed the duplicate `built-in` badge from the sources list row when the source kind already renders `built-in`
- removed the duplicate `built-in` badge from the source detail header for the same case

### After

image

### Before
CleanShot 2026-04-15 at 08 23 59@2x

## Why
The UI was rendering two adjacent `built-in` chips for built-in sources because it showed:
- a runtime badge when `runtime === true`
- a kind badge when `kind === "built-in"`

This PR keeps the runtime badge behavior for non-built-in runtime sources, but suppresses the extra chip when the kind already conveys the same label.

## Impact
Built-in sources now show a single `built-in` badge in the header and source list entry instead of a duplicated chip.

## Validation
- `bun run --cwd packages/react typecheck` in the main workspace
- verified locally with the `playwriter` CLI against `apps/local` at `http://127.0.0.1:4113/sources/built-in`

### DIFF
--- a/packages/react/src/pages/source-detail.tsx
+++ b/packages/react/src/pages/source-detail.tsx
@@ -149,7 +149,7 @@ export function SourceDetailPage(props: { namespace: string }) {
           <h2 className="truncate text-sm font-semibold text-foreground">
             {sourceData?.name ?? namespace}
           </h2>
-          {sourceData?.runtime && (
+          {sourceData?.runtime && sourceData?.kind !== "built-in" && (
             <Badge className="bg-muted text-muted-foreground">built-in</Badge>
           )}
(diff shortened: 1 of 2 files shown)


RhysSullivan:
Closing — the `kind !== "built-in"` check is a no-op (no plugin sets `kind: "built-in"`; the duplicate chip actually comes from `kind: "control"` on graphql/openapi static sources). The class of bug — two ad-hoc badge slots inferring 'is this a system source?' from different fields in two files — needs a shared `getSourceBadges()` helper, not another magic-string check. Will revisit holistically.

PR #261, opened 2026-04-15
Allow "Remote"-ing to Executor Web Dashboard

therealsamyak:
## What Changed

Added `--hostname` and `--exposed` CLI flags to `executor web`:

- `--hostname <addr>` changes the `Bun.serve` bind address (default: `127.0.0.1`)
- `--exposed` binds to `0.0.0.0` and disables the host allowlist check, allowing remote access from non localhost clients

## Why

When running `executor web` in Docker or on a remote machine (ex. trying to access over Tailscale), the server needs to bind to `0.0.0.0` instead of `127.0.0.1`, and the `isAllowedHost` check must be bypassed since non localhost `Host` headers get rejected with `403`.

## Impact

Two files changed:

- `apps/local/src/serve.ts`
  - Added `hostname` and `disableHostCheck` to `StartServerOptions`
  - Used in `Bun.serve()` and host check logic

- `apps/cli/src/main.ts`
  - Added `--hostname` and `--exposed` flags to the `web` command
  - Wired through to `startServer()`

Default behavior unchanged: `executor web` with no flags behaves identically to upstream.

## Validation

- `bun vitest run`
  - 141/143 tests passing
  - 2 pre existing failures in preset icon tests (unrelated)

- `bun run apps/cli/src/main.ts web --help`
  - Shows `--hostname` and `--exposed`

- `executor web --exposed`
  - Binds to `0.0.0.0`
  - Non localhost `Host` headers return `200`

- `executor web`
  - Binds to `127.0.0.1`
  - Non localhost `Host` headers return `403`
  
  
## Notes
This PR was vibe-coded with with Opencode + Oh-my-Openagent + GLM-5.1 / GLM-5-Turbo.

The `--hostname` flag might not be necessary. If you are going to host this on a custom domain / IP, you could instead use the `--exposed` flag, then containerize + reverse proxy, etc., instead of putting that burden on this repo / maintainers moving forward. 

I am not that familiar with networking stuff so correct me if I am wrong here...

Feel free to checkout the branch and push changes accordingly.

### DIFF
--- a/apps/cli/src/main.ts
+++ b/apps/cli/src/main.ts
@@ -111,11 +111,12 @@ const makeApiClient = (baseUrl: string) =>
 // Foreground session
 // ---------------------------------------------------------------------------
 
-const runForegroundSession = (input: { port: number; hostname: string; noCors: boolean }) =>
+const runForegroundSession = (input: { port: number; hostname: string; exposed: boolean }) =>
   Effect.gen(function* () {
-    const server = yield* Effect.promise(() => startServer({ port: input.port, hostname: input.hostname, disableHostCheck: input.noCors, embeddedWebUI }));
+    const hostname = input.exposed ? "0.0.0.0" : input.hostname;
+    const server = yield* Effect.promise(() => startServer({ port: input.port, hostname, disableHostCheck: input.exposed, embeddedWebUI }));
 
-    const baseUrl = `http://${input.hostname}:${server.port}`;
+    const baseUrl = `http://${hostname}:${server.port}`;
     console.log(`Executor is ready.`);
     console.log(`Web:     ${baseUrl}`);
     console.log(`MCP:     ${baseUrl}/mcp`);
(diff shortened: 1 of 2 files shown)


mrzmyr (commented):
## Claude Code Review (Opus 4.6)

Clean PR — the feature makes sense and the implementation is minimal. Two things I'd fix:

### 1. Display URL is broken when using `--exposed`

When `--exposed` sets hostname to `0.0.0.0`, the console prints `http://0.0.0.0:4788` which doesn't work in a browser. Should still display `localhost`:

```ts
const displayHost = hostname === "0.0.0.0" ? "localhost" : hostname;
const baseUrl = `http://${displayHost}:${server.port}`;
```

### 2. `--exposed` silently overrides `--hostname`

If someone passes `--hostname 10.0.0.5 --exposed`, the hostname is quietly ignored. These should either be mutually exclusive (error if both provided) or `--exposed` should only control `disableHostCheck` without touching hostname.

RhysSullivan:
Swapped it to:
- Dropped `--exposed`. It disabled the Host-header allowlist entirely, which removes the DNS-rebinding protection. If someone ran `executor web --exposed` on a network a browser could reach, any site they visited could rebind DNS and hit the executor API
- Kept `--hostname` (same semantics: bind address).
- Added `--allowed-host <host>` (repeatable). Appends to the built-in allowlist instead of bypassing it. Localhost/127.0.0.1 are always allowed.
- Added a startup warning when binding non-locally.
- Printed `baseUrl` uses `localhost` instead of `0.0.0.0` (the literal isn't a visitable URL).

PR #265, opened 2026-04-16
Add totalSources to tools.executor.sources.list() output

matiacone:
## Why

`tools.executor.sources.list()` returns a paginated list (default 12). 

When a user has more than 12 sources, the model mistakes that first page for the full list.

I hit this myself: a source I'd just registered was at alphabetical position 13, so `sources.list()` never returned it and I spent a while debugging a "missing" source that was there the whole time.

## What

Change the return shape from `ExecutorSourceListItem[]` to `{ sources, totalSources }`. `totalSources` is the count after filtering but before the limit is applied.

Computing it is free given the full filtered+counted list is already materialized before `.slice(0, limit)`.

### DIFF
--- a/packages/core/execution/src/tool-invoker.ts
+++ b/packages/core/execution/src/tool-invoker.ts
@@ -65,6 +65,11 @@ export type ExecutorSourceListItem = {
   readonly toolCount: number;
 };
 
+export type ExecutorSourceListResult = {
+  readonly sources: readonly ExecutorSourceListItem[];
+  readonly totalSources: number;
+};
+
 type SearchableTool = Pick<Tool, "id" | "sourceId" | "name" | "description">;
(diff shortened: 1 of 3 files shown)


packages/core/execution/src/tool-invoker.ts
```diff
@@ -65,6 +65,11 @@ export type ExecutorSourceListItem = {
   readonly toolCount: number;
 };
 
+export type ExecutorSourceListResult = {
+  readonly sources: readonly ExecutorSourceListItem[];
+  readonly totalSources: number;
```
mrzmyr:
```suggestion
  readonly total: number;
```

RhysSullivan:
Closed in favor of #492.

PR #270, opened 2026-04-17
fix(openapi): use Workers-native fetch for spec URLs

RhysSullivan:
Adding an OpenAPI source via URL hangs on production Cloudflare Workers.
Two compounding causes:

1. @apidevtools/swagger-parser fetches URLs via @apidevtools/json-schema-ref-parser's
   HTTP resolver, which goes through Node's http/https under nodejs_compat and
   stalls indefinitely on prod workerd. It also inlines every $ref on bundle(),
   deep-cloning shared schemas — a 16MB spec balloons to hundreds of MB and
   trips the 128MB Worker cap before it can finish.

2. addSpec wrapped parse + extract inside ctx.transaction, so the pool=1
   Hyperdrive Postgres connection held BEGIN open across the network fetch.

Changes:

- parse.ts: drop @apidevtools/swagger-parser entirely. Add fetchSpecText(url)
  that uses @effect/platform's HttpClient with a 20s Effect.timeout — pluggable
  via FetchHttpClient.layer (Workers-native) or a test-injected fetch. parse()
  is now JSON.parse + YAML fallback + 3.x validation; no ref resolution, because
  DocResolver (openapi-utils.ts) already resolves $refs lazily and
  normalizeOpenApiRefs preserves them through to the $defs table. Removed the
  manual resolveRefsInPlace walker — same deep-clone memory problem.

- plugin.ts: resolveSpecText(config.spec) runs with httpClientLayer provided;
  parse + extract run BEFORE ctx.transaction. Only the three storage writes
  (upsertSource, sources.register, definitions.register) stay inside the tx,
  so Hyperdrive's single connection isn't held during network I/O. detect() +
  previewSpec() get the same treatment.

- preview.ts / preview-oauth2.test.ts / real-specs.test.ts: wrap previewSpec
  with FetchHttpClient.layer where tests call it directly. real-specs.test.ts
  also strips the three in-memory-adapter addSpec tests — equivalent coverage
  lives in the cloud-app integration suite after this stack lands.

- package.json / bun.lock: drop @apidevtools/swagger-parser.

openapi test suite: 30/30 pass, 2.4s (was 12.4s with swagger-parser).

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -669,7 +669,6 @@
       "name": "@executor/plugin-openapi",
       "version": "0.0.1",
       "dependencies": {
-        "@apidevtools/swagger-parser": "^12.1.0",
         "@effect/platform": "catalog:",
         "@effect/platform-node": "catalog:",
         "@executor/config": "workspace:*",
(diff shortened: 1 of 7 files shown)


RhysSullivan:
* **#274** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/274?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#273** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/273?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#272** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/272?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#271** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/271?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#270** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/270?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/270?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #282, opened 2026-04-17
feat(sdk): typed errors propagate to one edge mapping (StorageError, InternalError, Telemetry)

RhysSullivan:
Cuts the codebase over from raw `Error` channels + per-handler `sanitize*`
functions to a typed-error model where storage failures propagate as
values to a single edge translator. Plugin code never imports a logger,
never calls `Sentry.captureException`, never wraps handlers — the edge
does that once.

## Storage layer (storage-core, storage-drizzle)

- `StorageError({ message, cause })` and `UniqueViolationError({ model? })`
  as `Data.TaggedError`. Cause carried as runtime data, never serialised.
- `DBAdapter`, `CustomAdapter`, `TypedAdapter` all declare typed `StorageFailure`
  channels — no more raw `Error` from the storage interface.
- `storage-drizzle` classifies pg `23505` (and sqlite equivalent) →
  `UniqueViolationError`; everything else → `StorageError({ cause })`.
- `storage-core/testing/memory` updated to match.
- `factory.ts` forwards typed errors; unknown-model becomes `StorageError`.

storage-core stays focused on data access — zero observability concerns.

## SDK (core/sdk)

- New `observability.ts`: `Telemetry` service tag + `InternalError({ traceId })`
  schema (HttpApi-shaped, status 500, opaque). The single 500 surface for
  the whole API.
- `executor.ts` captures Telemetry once at construction and builds
  `liftStorage` — the one translator that converts `StorageError` to
  captured `InternalError(traceId)` while passing `UniqueViolationError`
  through. Applied at every PluginCtx surface (ctx.storage, ctx.core.*,
  ctx.secrets.*, ctx.transaction) AND at the public Executor surface.
- `PluginCtx` declares `Effect<X, InternalError | UniqueViolationError>` for
  every infra method — plugin code never sees raw `StorageError`.
- `createExecutor`'s effect now requires `Telemetry` — hosts provide an
  implementation; tests use `Telemetry.NoOp`.
- `Telemetry` and `InternalError` re-exported from `@executor/sdk` for
  plugin / API consumers.

## API edge (core/api)

- Re-exports `Telemetry` + `InternalError` from `@executor/sdk` for HTTP
  wiring convenience.
- `observabilityMiddleware(api)` builds an `HttpApiBuilder.middleware`
  layer — defect safety net. Catches anything the typed channels missed,
  captures via Telemetry, returns a typed `InternalError(traceId)` body.
  Should rarely fire when the SDK is well-typed.

## MCP plugin (template for the other plugins)

- `McpExtensionFailure = McpOAuthError | McpConnectionError | McpToolDiscoveryError | McpInvocationError | InternalError`.
- `McpGroup.addError(...)` declares all five at the group level once —
  no per-endpoint addError, no per-plugin `McpInternalError`.
- Plugin domain errors get `HttpApiSchema.annotations({ status: 400 })` so
  the framework encodes them per-tag.
- `UniqueViolationError` is intentionally not in the public surface —
  plugins are expected to `Effect.catchTag` it and translate to their own
  user-facing error before exposing.
- (One pragmatic `as unknown as McpPluginExtension` cast inside
  `plugin.ts` because internal MCP SDK helpers still leak generic `Error`.
  TODO'd to clean up alongside the other plugin migrations.)

## Hosts

- Cloud: `apps/cloud/src/observability.ts` provides Sentry-backed
  `TelemetryLive`. `createScopedExecutor` provides it to `createExecutor`.
  Protected API gets the edge middleware.
- Local CLI: provides `Telemetry.NoOp` (storage failures still die
  cleanly via Effect's standard crash path).
- Marketing detect, examples/all-plugins, test harness, promise SDK:
  same `Telemetry.NoOp` provision.

## Tests

- Every `createExecutor` call site in tests pipes through
  `Effect.provide(Telemetry.NoOp)`.
- One existing executor test was checking error-message contents on a
  collision ("collides with a static source"); under the new model that's
  captured to telemetry and surfaced as `InternalError`. Test now asserts
  `_tag === "InternalError"`.

## Notes

- `notes/error-handling.md` documents the model, where each piece lives,
  and the anti-patterns to avoid (`Effect.orDie` at handler boundaries,
  per-plugin `*InternalError` types, `sanitize*` helpers, plugin SDK
  importing `Sentry.captureException` directly).
- Other plugins (OpenAPI, GraphQL, Google Discovery, OnePassword) still
  use the old surface and need the same MCP-style migration. Out of
  scope for this PR — review the model on MCP first.


RhysSullivan:
* **#282** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/282?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/282?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/core/sdk/src/executor.ts
```diff
@@ -558,17 +584,23 @@ export const createExecutor = <
...
-            new Error(`Secret provider "${target.key}" is read-only`),
+            new StorageError({
+              message: `Secret provider "${target.key}" is read-only`,
+              cause: undefined,
+            }),
           );
         }
 
-        yield* target.set(input.id, input.value);
+        yield* wrapProviderError(target.set(input.id, input.value));
```
RhysSullivan: wrap provider error is weird, looks like this is being used to mask new Error from other poitns we havent upated, which we'll just go through and fix those rather than doing this so we get the full error info

apps/local/src/server/main.ts
```diff
@@ -79,11 +78,11 @@ const closeServerHandlers = async (handlers: ServerHandlers): Promise<void> => {
 
 export const createServerHandlers = async (): Promise<ServerHandlers> => {
   const executor = await getExecutor();
-  const engine = createExecutionEngine({ executor, codeExecutor: makeQuickJsExecutor() });
+  const engine = createExecutionEngine({ executor });
 
   const pluginExtensions = Layer.mergeAll(
     Layer.succeed(OpenApiExtensionService, executor.openapi),
-    Layer.succeed(McpExtensionService, executor.mcp),
+    Layer.succeed(McpExtensionService, withStorageCapture(executor.mcp)),
```
RhysSullivan: specifically 'withStorageCapture' is odd, think more generic about with capture or something like that and applying it to all errors

apps/local/src/server/executor.ts
```diff
@@ -153,6 +153,10 @@ const createLocalExecutorLayer = () => {
         createdAt: new Date(),
       });
 
+      // Local app doesn't wire `ErrorCapture` — storage failures still
```
RhysSullivan: just log to console

PR #286, opened 2026-04-18
marketing: wrap Astro Worker with otel-cf-workers + Sentry

RhysSullivan:
## Summary

Mirror `apps/cloud`'s observability wiring on the marketing Worker: traces go to Axiom via `@microlabs/otel-cf-workers`, errors to Sentry via `@sentry/cloudflare`. Both workers write to the same Axiom dataset (`executor-cloud`) and are distinguished by `service.name` (`executor-marketing` vs `executor-cloud`).

## Approach

The Astro Cloudflare adapter emits `dist/server/entry.mjs` as the Worker entry and there is no hook to substitute that entry, so we added a **post-build shim** step (`apps/marketing/scripts/wrap-worker.ts`):

1. Rename `dist/server/entry.mjs` -> `dist/server/astro-entry.mjs`.
2. Write a small ESM shim that imports `./astro-entry.mjs` and wraps its default export with `Sentry.withSentry(sentryOptions, instrument({ fetch: astroHandler.fetch }, otelConfig))` (other top-level handlers like `scheduled`/`queue` are preserved via spread).
3. Bundle the shim with `Bun.build` (Sentry + otel-cf-workers inlined — Vite has already run, so anything new here needs to ship self-contained). The renamed Astro entry stays as an external relative import so its own chunk graph keeps working under wrangler's `no_bundle: true` rule.
4. Defensively ensure `compatibility_flags` in `dist/server/wrangler.json` includes `nodejs_compat`.

`sentryOptions` matches `apps/cloud/src/server.ts` verbatim: `skipOpenTelemetrySetup: true`, `tracesSampleRate: 0`, `enableLogs: true`, `sendDefaultPii: true`. Env reads are direct `process.env.*` since marketing has no `@executor/env` package. `skipOpenTelemetrySetup: true` is load-bearing — otel-cf-workers owns the global TracerProvider; Sentry's OTEL shim would clobber it and break every request with `global tracer is not of type WorkerTracer`.

We deliberately do NOT use `instrumentDO` or `Sentry.instrumentDurableObjectWithSentry` (marketing has no DOs, and those wrappers have known `this`-binding issues on WorkerTransport anyway).

## Why this approach vs alternatives

- **Astro integration hook**: Astro's build lifecycle has no way to replace the adapter's generated entry file — the adapter's `astro:build:setup` already owns the Rollup input. A post-build wrap is the cleanest intercept point.
- **Edit the generated entry in place**: Brittle. Astro's entry imports hashed chunks via helpers; rewriting that file would tie us to adapter internals. Isolating our changes to a new shim file and leaving Astro's output untouched is safer.
- **Import-from-source wrapping**: Would require running Astro's Vite pass through our config, not realistic.

## Testing

- [x] `bun run build` in `apps/marketing` succeeds and emits `dist/server/entry.mjs` (~390 KB bundle with Sentry + otel-cf-workers inlined) plus `dist/server/astro-entry.mjs` (unchanged Astro output, relative-imported by the shim).
- [x] `bun run typecheck` at the repo root passes (34/34 tasks green).
- [x] `bun run lint` at the repo root passes.
- [ ] **Deploy** (user): set `AXIOM_TOKEN`, optional `AXIOM_DATASET` (defaults to `executor-cloud`), and `SENTRY_DSN` as secrets on the `executor-marketing` Worker, then `bun run deploy` from `apps/marketing`.
- [ ] **Verify** (user): trigger a request, confirm a trace shows up in the Axiom `executor-cloud` dataset under `service.name = "executor-marketing"`, throw a test error and confirm it lands in the `node-cloudflare-workers` Sentry project.

### DIFF
--- a/apps/marketing/package.json
+++ b/apps/marketing/package.json
@@ -4,9 +4,9 @@
   "type": "module",
   "scripts": {
     "dev": "portless run --no-tls --name executor-marketing astro dev",
-    "build": "astro build",
+    "build": "astro build && bun run scripts/wrap-worker.ts",
     "preview": "astro preview",
-    "deploy": "astro build && npx wrangler deploy --config dist/server/wrangler.json",
+    "deploy": "astro build && bun run scripts/wrap-worker.ts && npx wrangler deploy --config dist/server/wrangler.json",
     "astro": "astro",
     "typecheck": "tsgo --noEmit",
     "typecheck:slow": "tsc --noEmit"
@@ -17,6 +17,9 @@
     "@executor/plugin-graphql": "workspace:*",
     "@executor/plugin-openapi": "workspace:*",
     "@executor/sdk": "workspace:*",
+    "@microlabs/otel-cf-workers": "^1.0.0-rc.52",
+    "@opentelemetry/api": "~1.9.0",
+    "@sentry/cloudflare": "^10.49.0",
     "@tailwindcss/vite": "^4.2.2",
     "astro": "^6.1.3",
     "effect": "^3.14.16",
--- a/apps/marketing/scripts/wrap-worker.ts
+++ b/apps/marketing/scripts/wrap-worker.ts
@@ -1,0 +1,181 @@
+#!/usr/bin/env bun
+/**
+ * Post-build step: wraps the Astro Cloudflare adapter output with
+ * otel-cf-workers + Sentry Cloudflare instrumentation.
+ *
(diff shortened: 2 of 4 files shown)


RhysSullivan:
Closing — the post-build shim adds maintenance surface for marginal signal on a static marketing site. Errors/traffic visible through executor-cloud's proxy span (Homepage dashboard) + CF Workers Observability is sufficient.

PR #288, opened 2026-04-18
cloud: port /mcp + /.well-known to Effect HttpApp

RhysSullivan:
## Summary
- Replaces `apps/cloud/src/mcp.ts`'s plain async handler with an Effect HTTP app served via `@effect/platform`'s `HttpApp.toWebHandler`. JWT verification is an `McpAuth` service + layer, span annotation uses `Effect.annotateCurrentSpan` (nests under the otel-cf-workers fetch span), and a single `Effect.catchAllCause` at the boundary replaces three scattered try/catches.
- DO dispatch stays imperative at the very edge: `stub.init()` and `stub.handleRequest()` are wrapped in `Effect.promise(...)` and the returned `Response` goes through `HttpServerResponse.raw(...)`. The platform's `toWeb` conversion recognises `body.body instanceof Response` and returns it as-is, so the SSE `ReadableStream` body passes through unchanged.
- `start.ts` now calls `mcpFetch(request)`; the handler returns `null` for non-MCP paths so TanStack Start keeps routing.

## Before / after sketch

Before — `handleMcpRequest` is plain async with imperative branching:

```ts
export const handleMcpRequest = async (request: Request): Promise<Response | null> => {
  const url = new URL(request.url);
  if (request.method === "OPTIONS" && isMcpPath(url.pathname)) return corsPreflight(...);
  if (url.pathname === "/.well-known/oauth-protected-resource") return protectedResourceMetadata();
  if (url.pathname === "/.well-known/oauth-authorization-server") return authorizationServerMetadata();
  if (url.pathname !== "/mcp") return null;

  const token = await verifyBearerToken(request);
  if (!token) return unauthorized();
  switch (request.method) { /* ... */ }
};
```

After — the handler is an Effect that reads `HttpServerRequest`, converted to a fetch handler via `HttpApp.toWebHandler`:

```ts
const mcpApp = Effect.gen(function* () {
  const httpRequest = yield* HttpServerRequest.HttpServerRequest;
  const request = httpRequest.source as Request;
  // ... route classification, CORS, OAuth metadata, auth, dispatch ...
}).pipe(
  Effect.withSpan("mcp.request"),
  Effect.catchAllCause((cause) => /* Sentry + jsonRpcError 500 */),
);

const rawMcpFetch = HttpApp.toWebHandler(
  mcpApp.pipe(Effect.provide(Layer.mergeAll(McpAuthLive, TelemetryLive))),
);

export const mcpFetch = async (request: Request): Promise<Response | null> =>
  classifyPath(new URL(request.url).pathname) === null ? null : rawMcpFetch(request);
```

## Preserved exactly
- CORS preflight headers on OPTIONS, including `access-control-expose-headers: mcp-session-id`
- 401 body `{ error: "unauthorized" }` + `www-authenticate: Bearer resource_metadata="..."` on unauthenticated /mcp
- 403 + JSON-RPC `-32001` for missing `organizationId` in verified token
- 405 + JSON-RPC `-32001` for wrong method
- 500 + JSON-RPC `-32603` for internal errors (caught at the Effect boundary, sent to Sentry)
- `/.well-known/*` stays public (no auth)
- DO streaming responses pass through intact via `HttpServerResponse.raw(response)` — `toWeb` unwraps the inner `Response` directly when `body.body instanceof Response`, so the SSE `ReadableStream` body is never re-read
- `mcp.*` / `mcp.auth.*` span attributes the pre-port code always set are still captured via `Effect.annotateCurrentSpan`; the richer fingerprint capture on `rs/mcp-do-shared-layer` slots into the same `annotateMcpRequest` helper when that branch lands (noted in a comment in the file).

## Test plan
- [x] `bun run typecheck` in `apps/cloud` — passes
- [x] `bun run typecheck` at repo root (turbo across all packages) — passes
- [x] `npx vitest run` in `apps/cloud` (workerd pool) — 5 files, 27 tests passing
- [x] `npx vitest run --config vitest.node.config.ts` in `apps/cloud` — 4 files, 23 tests passing, including the full `mcp-session.e2e.node.test.ts` end-to-end suite
- [ ] After merge: redeploy and verify in Axiom that /mcp requests still get the `mcp.request.*` + `mcp.auth.*` span attributes and that DO span nesting still looks right. Same MCP-client reconnect dance as before — clients will lose their existing session on the deploy and need to re-init.

### DIFF
--- a/apps/cloud/src/mcp.ts
+++ b/apps/cloud/src/mcp.ts
@@ -108,25 +108,155 @@ export const McpAuthLive = Layer.succeed(McpAuth, {
 });
 
 // ---------------------------------------------------------------------------
-// Span annotation
+// Client fingerprint capture
 // ---------------------------------------------------------------------------
 // Annotates the Effect span (which nests under the otel-cf-workers fetch
-// span) with the minimum we always know about an MCP request. Richer
-// fingerprint capture (parsed JSON-RPC body, whitelisted headers, CF meta)
-// lives on rs/mcp-do-shared-layer and slots in here when that branch lands.
+// span) with everything we can learn about a connecting MCP client: the
+// parsed JSON-RPC body, whitelisted request headers, CF request metadata,
+// and verified-JWT claims. Lets us compare how each client (Claude Code,
+// Claude.ai web, ChatGPT, custom scripts, ...) actually reports over the
+// wire. Runs before dispatch so unauthorized requests still get fingerprinted.
+// ---------------------------------------------------------------------------
+
+type CfRequestMetadata = {
+  country?: string;
+  city?: string;
+  region?: string;
+  timezone?: string;
+  asn?: number;
+  asOrganization?: string;
+  tlsVersion?: string;
+  tlsCipher?: string;
+  httpProtocol?: string;
+  colo?: string;
+};
+
+const getCfMeta = (request: Request): CfRequestMetadata =>
+  ((request as unknown as { cf?: CfRequestMetadata }).cf ?? {}) as CfRequestMetadata;
+
+const HEADERS_TO_DUMP = [
+  "accept",
+  "accept-encoding",
+  "accept-language",
+  "cache-control",
+  "content-type",
+  "mcp-protocol-version",
+  "origin",
+  "referer",
+  "sec-fetch-dest",
+  "sec-fetch-mode",
+  "sec-fetch-site",
+  "user-agent",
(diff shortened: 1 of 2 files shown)


RhysSullivan:
Superseded by #289, which folds this PR's commits plus the multi-request workerd coverage, real-port Miniflare e2e, and MCP elicitation round-trip into a single PR targeting main.

PR #343, opened 2026-04-20
[codex] mcp: name oauth secrets after source display name

heyglassy:
## What changed
MCP OAuth token secrets now use the source/session display name instead of fixed generic labels.

## Why
The MCP flow was storing secrets as `MCP OAuth Access Token` and `MCP OAuth Refresh Token`, which made multiple MCP connections hard to distinguish in the secrets UI.

### Before 
CleanShot 2026-04-20 at 10 21 26@2x

### After
CleanShot 2026-04-20 at 10 25 27@2x

## Root cause
The MCP start/complete OAuth flow did not carry the source display name through the persisted OAuth session, so token writes had no source-specific label available and fell back to hardcoded names.

## Validation
- `bunx vitest run src/sdk/plugin.test.ts`
- `bun run typecheck`

### DIFF
--- a/packages/plugins/mcp/src/api/group.ts
+++ b/packages/plugins/mcp/src/api/group.ts
@@ -105,6 +105,7 @@ const NamespacePayload = Schema.Struct({
 });
 
 const StartOAuthPayload = Schema.Struct({
+  name: Schema.String,
   endpoint: Schema.String,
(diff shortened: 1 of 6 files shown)


RhysSullivan:
closed in favor of 'connections' concept

PR #357, opened 2026-04-21
skills: add @executor/plugin-skills with per-source attachment

RhysSullivan:
## Summary
- New `@executor/plugin-skills` package. A skill is a markdown body plus a name + description; the plugin registers each one as a static tool whose handler returns the body. Discovery is the existing `tools.list({ query })`; loading is `tools.invoke(id)`. No new primitives, no core SDK changes.
- Exports `toStaticSkill(skill): StaticToolDecl` so any plugin can declare its own skills in its own `staticSources` — the global `skillsPlugin` is just one consumer of the same helper.
- `@executor/plugin-openapi` ships its first skill (`openapi.adding-a-source`, a preview → resolve-auth → addSource walkthrough) and registers it on its own source. `tools.list({ sourceId: "openapi" })` returns `openapi.previewSpec`, `openapi.addSource`, and `openapi.adding-a-source` together — the playbook sits literally next to the tools it documents.
- `apps/local` wires `skillsPlugin()` (empty, reserved for cross-cutting / user-authored skills) — openapi owns its own skills directly.

## Safety: no user-typed secrets in the agent's context
The skill body was reworded so it never instructs the agent to accept a secret value in chat and call `secrets.set` with it. That path routes the raw value through the LLM before it gets stored, which is exactly the leak we want to avoid. The new rule: pick an id, reference it in step 3, the user provisions the value out of band. A test pins the policy so a future edit can't silently regress.

## Context: MCP Skills Over MCP WG
The first-class-primitive draft (SEP-2076: `skills/list` + `skills/get`) was closed 2026-02-24. The Skills Over MCP WG (promoted from Interest Group four days before opening this PR) has accepted **skills-as-resources** with a `skill://<skill-path>/SKILL.md` URI scheme. Because our static-tool system is effectively in-process MCP, the attachment point should match: skills live under the sourceId of the server/plugin that ships them. That's why openapi owns its own skill now; when the SEP stabilizes, renaming "static tool with markdown body" → "static skill under `skill://` URI" is a shape change inside the same place. Full background and the deferred refactor plan: [`notes/skills.md`](notes/skills.md).

## Test plan
- [x] `tools.list({ sourceId: "openapi" })` returns previewSpec, addSource, and adding-a-source together
- [x] `tools.invoke("openapi.adding-a-source")` returns the markdown body
- [x] Skill body mentions both `openapi.previewSpec` and `openapi.addSource` (catches drift if a tool is renamed)
- [x] Skill body does not contain "store it via `secrets.set`" and does contain "Never accept a secret value in-chat"
- [x] Duplicate skill ids throw at plugin construction time
- [x] Empty skill list works (the apps/local case)
- [x] 49 tests passing across `@executor/plugin-skills` and `@executor/plugin-openapi`; typecheck clean in both + `apps/local`

## Follow-ups (not in this PR)
- `StaticSource { tools, skills }` as a sibling field, once the MCP Skills Extension SEP stabilizes.
- Secure secret capture (MCP elicitation wired to a UI that bypasses the model). The skill forbids in-chat values but doesn't yet have a UX to replace it.

### DIFF
--- a/packages/plugins/openapi/src/sdk/skills.ts
+++ b/packages/plugins/openapi/src/sdk/skills.ts
@@ -1,12 +1,14 @@
 import type { Skill } from "@executor/plugin-skills";
 
-// Skills shipped alongside the OpenAPI plugin. Consumers wire them in by
-// passing this array to `skillsPlugin({ skills: [...openapiSkills] })`.
-// Every skill id is prefixed `openapi.` so a catch-all `tools.list({ query: "openapi" })`
-// surfaces it next to the real openapi tools.
+// Skills shipped by the OpenAPI plugin. Registered into the plugin's
+// own `staticSources` (see `plugin.ts`) — NOT passed to the global
+// `skillsPlugin`. Living under sourceId `openapi` means the full tool
+// id is `openapi.<skill-id>`, right next to `openapi.previewSpec` etc.
+// Skill ids therefore omit any `openapi.` prefix; the source id is the
+// attachment point.
 export const openapiSkills: readonly Skill[] = [
   {
-    id: "openapi.adding-a-source",
+    id: "adding-a-source",
     description:
       "How to add an OpenAPI spec as a source — preview, resolve auth, then addSpec",
     body: `# Adding an OpenAPI source
--- a/packages/plugins/openapi/src/sdk/skills.ts
+++ b/packages/plugins/openapi/src/sdk/skills.ts
@@ -1,0 +1,69 @@
+import type { Skill } from "@executor/plugin-skills";
+
+// Skills shipped alongside the OpenAPI plugin. Consumers wire them in by
+// passing this array to `skillsPlugin({ skills: [...openapiSkills] })`.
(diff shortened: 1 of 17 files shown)


packages/plugins/openapi/src/sdk/skills.ts
```diff
@@ -0,0 +1,69 @@
...
+
+**Why this step:** the preview tells you whether the spec needs
+credentials, and which scheme to use. Do not skip it — \`addSpec\` will
+fail at invoke time if required auth isn't wired.
+
+## 2. Resolve authentication
+
+Look at the preview's \`securitySchemes\`:
+
+- **API key / bearer** — ask the user for the value, then store it via
```
RhysSullivan: LLMs should not be listing secrets from users.

mrzmyr:
very nice PR!

RhysSullivan:
Closing this as stale. The current skills direction is being tracked in Linear via EXE-45 and EXE-46, and the old branch is no longer a good base after the recent platform changes.

PR #366, opened 2026-04-22
openapi: dispatch non-JSON request bodies by content type

RhysSullivan:
## Summary

Invoke previously had only two real body branches — JSON and \`String(bodyValue)\` with whatever content type the spec declared. Multipart uploads, XML bodies, text/plain, and octet-stream byte uploads all collapsed into either \`[object Object]\` on the wire or the wrong framing.

New \`applyRequestBody\` dispatches on the spec-declared content type first, JS body type second:

- \`multipart/form-data\` → \`bodyFormDataRecord\` (primitives pass through, \`Uint8Array\` / \`ArrayBuffer\` / \`Buffer\` auto-Blob, nested objects JSON-stringify, arrays supported). Pre-built \`FormData\` accepted via \`bodyFormData\`.
- \`application/xml\`, \`text/xml\`, \`*+xml\` → \`bodyText\` passthrough when string, JSON.stringify fallback so the wire never carries \`[object Object]\`.
- \`text/*\` → text passthrough.
- \`application/octet-stream\` → \`bodyUint8Array\` for any byte shape; string fallback.
- JSON and form-urlencoded paths preserved byte-for-byte.

## Test plan

- [ ] \`non-json-body.test.ts\` — tiny node echo servers assert wire-level content-type + body for multipart (mixed primitives), \`application/xml\` (string), \`text/xml\` (object), \`text/plain\` (string), \`application/octet-stream\` (\`Uint8Array\`).
- [ ] \`@executor/plugin-openapi\` 42/42 (37 prior + 5 new).


RhysSullivan:
Follow-up commit (`70a2aba4`) closes the spec-compliance gaps I flagged when auditing against swagger-client's implementation. Now PR covers:

**From the original commit** — dispatch non-JSON bodies by content type (multipart, xml, text, octet-stream) instead of falling through to `String(bodyValue)`.

**New in 70a2aba4**:

1. **Multi-content requestBody**. Specs declaring more than one media type per operation no longer silently collapse to JSON-first. `preferredContent` follows swagger-client's first-declared-wins (spec author's order). When multiple are declared, the tool's input schema exposes a `contentType` enum so the model can pick; default stays the first-declared.

2. **`encoding` object (§4.8.15) extracted and threaded through invoke**. New `MediaBinding` and `OperationRequestBody.contents` preserve every declared media type with its own schema + encoding.

3. **Multipart `encoding.contentType` honored**. Parts with a declared per-part content type (common case: JSON metadata alongside a binary file) now ship wrapped as a typed `Blob`, so the runtime framer emits the per-part `Content-Type` header.

4. **Form-urlencoded `style` / `explode` / `allowReserved` honored per-field**. `style: form, explode: true` (default) repeats the key for arrays; `explode: false` comma-joins (or pipe / space per `pipeDelimited` / `spaceDelimited`). `style: deepObject` emits `field[subkey]=value` with brackets URL-encoded — matches swagger-client's wire format.

Response-side content negotiation kept its JSON-first preference via the new `preferredResponseContent`.

Tests: 37 prior + 12 new = 49/49 green. Affected package tests (`@executor/plugin-openapi` 49/49, `@executor/plugin-mcp` 22/22, `@executor/sdk` 85/85) all pass.

PR #396, opened 2026-04-24
feat(sdk): add ExecutionStore backed by DBAdapter

aryasaatvik:
## Stack

**Root of a 5-PR stack — merge first.** Cross-fork GitHub PRs can't use a branch on a contributor fork as a base, so these five PRs display as independent in the UI. The real dependency chain is the commit graph.

| # | PR | Purpose |
|---|---|---|
| 1 | **#396 ← you are here** | `feat(sdk)`: ExecutionStore backed by DBAdapter |
| 2 | #398 | `feat(execution)`: persist engine runs + tool calls |
| 3 | #399 | `feat(execution)`: trigger propagation (CLI/HTTP/MCP) |
| 4 | #400 | `feat(apps)`: execution tables in drizzle schemas |
| 5 | #401 | `feat(api)`: /executions list/get/tool-calls endpoints |

Each subsequent PR's diff includes this one's commits until this lands — after that the diffs cleanly separate.

---

## Summary

Adds execution history persistence as a first-class concern of the SDK, wiring three tables into `coreSchema` (`execution`, `execution_interaction`, `execution_tool_call`) and exposing an `ExecutionStore` service on `executor.executions`. Generic — every backend that implements the `DBAdapter` contract (sqlite, postgres, memory) inherits the store for free.

## What this PR adds

- **`core-schema.ts`**: three new tables with indexes on `scope_id`, `execution_id`, `tool_path`, `trigger_kind`, `created_at` to back runs UI facets + timeline queries.
- **`ids.ts`**: branded `ExecutionId`, `ExecutionInteractionId`, `ExecutionToolCallId`.
- **`executions.ts`**: `Execution` / `ExecutionInteraction` / `ExecutionToolCall` Schema classes, status enums, create/update/list/filter input types, and the `ExecutionStore` `Context.Tag`.
- **`execution-store.ts`**: `makeExecutionStore(core)` — adapter-backed implementation. Cursor pagination (`cursor.ts`), filter predicates (status/trigger/tool-glob/time/code/hadElicitation), and meta aggregation (facets + chart buckets).
- **`executor.ts`**: constructs the store once per executor, exposed via `executor.executions`.
- **`executions.test.ts`**: round-trip + lifecycle coverage against the in-memory adapter.

## What this PR does NOT do

Follow-up PRs in the stack will:
1. Wire the engine to call `executor.executions.create/update/record*/finish*` on execute/resume/tool-invoke boundaries (#398).
2. Thread trigger context from CLI and MCP hosts (#399).
[...]

PR #398, opened 2026-04-24
feat(execution): persist engine runs + tool calls via ExecutionStore

aryasaatvik:
## Stack

**Depends on #396 — merge that first.** Cross-fork GitHub PRs can't use a branch on a contributor fork as a base, so these five PRs display as independent in the UI. The real dependency chain is the commit graph.

| # | PR | Purpose |
|---|---|---|
| 2 | **#398 ← you are here** | `feat(execution)`: persist engine runs + tool calls |
| 3 | #399 | `feat(execution)`: trigger propagation (CLI/HTTP/MCP) |
| 5 | #401 | `feat(api)`: /executions list/get/tool-calls endpoints |

Until #396 lands, this diff includes its commits. After #396 merges, this diff shrinks to just the engine changes below.

---

## Summary

Wires the `ExecutionStore` added in #396 into the Effect-native engine. Every `execute()` / `executeWithPause()` / `resume()` call now writes execution + tool-call + interaction rows through whichever `DBAdapter` backs the SDK — sqlite, postgres, memory, anything else that implements the contract gets history for free.

## What ships in this PR (delta beyond #396)

**Engine API:**
- `ExecutionTrigger` type + new `trigger?` option on `execute` / `executeWithPause`. Callers attribute runs (`cli`, `http`, `mcp`, …); kind + optional meta blob persist on the row.
- Execution id is a `crypto.randomUUID()` minted at engine entry and reused as `PausedExecution.id`, so caller-visible ids and DB row ids are the same value.

**Recording:**
- `makeRecordingInvoker` wraps the `SandboxToolInvoker` passed to the code executor: each `invoke` writes a tool-call row (`running → completed | failed`) with `durationMs`. Storage failures are ignored so bookkeeping can never fail the tool call itself.
- `persistTerminalState` runs once on fiber success/failure and writes final `status`, `resultJson`, `errorText`, `logsJson`, `toolCallCount`, `completedAt`.
- Elicitation lifecycle (both inline + pausable paths): execution transitions to `waiting_for_interaction`, pending `execution_interaction` row created; on resume the row is resolved (or cancelled if `action === "cancel"`) before the fiber is unblocked.
- `toolCallCounters` keeps the same `Ref` across pause/resume so the final count is accurate even for multi-pause runs.

## Test plan

- [x] `bun x vitest run` in `@executor/execution` — 15/15 tests pass (10 existing + 5 new in `engine-persistence.test.ts`).
- [x] `bun x tsc --noEmit` — zero type errors.
- [x] `bun x vitest run` in `@executor/sdk` — 97/97 tests still pass.

### New test coverage (`engine-persistence.test.ts`)

1. Completed run records status + toolCallCount + tool call rows.
2. Errored result → status=`failed`, errorText captured.
3. Interaction lifecycle wired cleanly.
4. Trigger kind + meta persist on the row.
5. Failed tool call records status=`failed` with errorText.

### DIFF
--- a/packages/core/execution/src/engine-persistence.test.ts
+++ b/packages/core/execution/src/engine-persistence.test.ts
@@ -1,0 +1,270 @@
+import { describe, expect, it } from "@effect/vitest";
+import { Effect } from "effect";
+
+import {
+  createExecutor,
+  definePlugin,
+  ElicitationResponse,
+  ExecutionId,
+  makeTestConfig,
+  type ElicitationHandler,
+} from "@executor/sdk";
+import { CodeExecutionError } from "@executor/codemode-core";
+import type {
+  CodeExecutor,
+  ExecuteResult,
+  SandboxToolInvoker,
+} from "@executor/codemode-core";
+
+import { createExecutionEngine } from "./engine";
+
+// ---------------------------------------------------------------------------
+// Stub CodeExecutor that drives the invoker + elicitation handler from a
+// fixed script. Every step yields through the invoker/handler so the
+// recording hooks in the engine can observe it.
+// ---------------------------------------------------------------------------
+
+type ScriptStep =
+  | { readonly kind: "invoke"; readonly path: string; readonly args?: unknown }
+  | { readonly kind: "elicit"; readonly message: string };
+
+const makeScriptedExecutor = (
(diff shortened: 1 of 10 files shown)


aryasaatvik:

PR #399, opened 2026-04-24
feat(execution): propagate trigger context from CLI, HTTP, and MCP hosts

aryasaatvik:
## Stack

**Depends on #398 → #396 — merge those first.** Cross-fork GitHub PRs can't use a branch on a contributor fork as a base, so these five PRs display as independent in the UI. The real dependency chain is the commit graph.

| # | PR | Purpose |
|---|---|---|
| 2 | #398 | `feat(execution)`: persist engine runs + tool calls |
| 3 | **#399 ← you are here** | `feat(execution)`: trigger propagation (CLI/HTTP/MCP) |
| 5 | #401 | `feat(api)`: /executions list/get/tool-calls endpoints |

Until #398 and #396 land, this diff includes their commits. After both merge, this diff shrinks to the trigger-wiring changes.

---

## Summary

Flows the `trigger: { kind, meta }` option the engine added in #398 end-to-end, so the execution row's `trigger_kind` column reflects the surface that actually kicked off the run. This is what the runs UI will facet on.

## What ships in this PR

**HTTP API (`packages/core/api`):**
- `/executions` POST now declares `x-executor-trigger` as an optional header in the schema.
- Handler reads it (defaulting to `"http"`) and passes it as the engine's trigger option.

**MCP host (`packages/hosts/mcp`):**
- Explicit `trigger: { kind: "mcp" }` on `engine.execute` (inline elicitation path) and `engine.executeWithPause` (paused flow).

**CLI:** already sends `x-executor-trigger: cli` from the merge commit — this PR just makes the server side honor it.

## Engine robustness improvement

While wiring this up, the MCP stdio integration test surfaced a subtle bug: `Effect.ignore` only catches **typed failures**. `storage-drizzle` throws a synchronous `Error` when a model isn't in the schema, which Effect promotes to a **defect** — and defects were bypassing \`ignore\`, leaking the raw error text into tool results.

This PR introduces a `silent` helper (`Effect.catchAllCause(() => Effect.void)`) and swaps every execution-history bookkeeping call to use it. Net effect: misconfigured storage now means no row gets persisted — the execution itself still succeeds and returns its real result.

Before: `[storage-drizzle] unknown model "execution" — not found in db._.fullSchema…` leaked to the MCP `execute` tool result text.
After: the result is `4` as expected, recording silently fails until the app's drizzle schema gets the tables (#400).

## Test plan

- [x] `bun x vitest run` in `@executor/execution` — 15/15.
- [x] `bun x vitest run` in `@executor/sdk` — 97/97.
- [x] `bun x vitest run` in `@executor/api` — 4/4.
- [x] `bun x vitest run` in `@executor/hosts/mcp` — 23/23 including the previously-failing `stdio-integration.test.ts > execute tool returns result over stdio transport`.
- [x] `bun x tsc --noEmit` clean across the stack.

### DIFF
--- a/apps/cli/src/main.ts
+++ b/apps/cli/src/main.ts
@@ -394,6 +394,7 @@ const executeCode = (input: {
       payload: {
         code: input.code,
       },
+      headers: { "x-executor-trigger": "cli" },
     });
 
     if (response.status === "paused") {
--- a/packages/core/api/src/executions/api.ts
+++ b/packages/core/api/src/executions/api.ts
@@ -11,6 +11,16 @@ const ExecuteRequest = Schema.Struct({
   code: Schema.String,
 });
 
+/**
+ * Optional header naming the surface that triggered this execution —
+ * `"cli"`, `"http"`, `"mcp"`, etc. Persisted on the execution row so
+ * the runs UI can facet by trigger kind. Defaults to `"http"` when
+ * absent.
+ */
+const ExecuteHeaders = Schema.Struct({
+  "x-executor-trigger": Schema.optional(Schema.String),
+});
+
 const CompletedResult = Schema.Struct({
   status: Schema.Literal("completed"),
   text: Schema.String,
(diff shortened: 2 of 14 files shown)


aryasaatvik:

PR #400, opened 2026-04-24
feat(apps): add execution tables to local + cloud drizzle schemas

aryasaatvik:
## Stack

**Depends on #399 → #398 → #396 — merge those first.** Cross-fork GitHub PRs can't use a branch on a contributor fork as a base, so these five PRs display as independent in the UI. The real dependency chain is the commit graph.

| # | PR | Purpose |
|---|---|---|
| 1 | #396 | `feat(sdk)`: ExecutionStore backed by DBAdapter |
| 2 | #398 | `feat(execution)`: persist engine runs + tool calls |
| 3 | #399 | `feat(execution)`: trigger propagation (CLI/HTTP/MCP) |
| 4 | **#400 ← you are here** | `feat(apps)`: execution tables in drizzle schemas |
| 5 | #401 | `feat(api)`: /executions list/get/tool-calls endpoints |

Until #399, #398, and #396 land, this diff includes their commits. After all three merge, this diff shrinks to just the schema + migration files.

---

## Summary

Adds the three `execution*` tables to both app drizzle schemas (sqlite + postgres), plus the `drizzle-kit generate` output, so `executor.executions` writes land on disk in real deployments.

Until this PR, persistence silently no-op'd — `storage-drizzle` throws on unknown models, and the engine's `silent` wrapper (added in #399) absorbs the defect. Tests against the in-memory adapter pass (that's what the earlier PRs verify), but nothing gets persisted in a production setup.

## What ships in this PR

**`apps/local/src/server/executor-schema.ts`** (sqlite):
- `execution`: `[scope_id, id]` PK; indexes on `scope_id`, `status`, `trigger_kind`, `created_at`.
- `execution_interaction`: standalone `id` PK (tenant isolation flows through parent execution); indexes on `execution_id`, `status`.
- `execution_tool_call`: standalone `id` PK; indexes on `execution_id`, `tool_path`, `namespace`.

**`apps/cloud/src/services/executor-schema.ts`** (pg-core): mirror with `bigint` for epoch-ms columns (`started_at`, `completed_at`, `duration_ms`, `tool_call_count`) and `timestamp` for Date columns (`created_at`, `updated_at`).

**Generated migrations:**
- `apps/local/drizzle/0004_fancy_red_wolf.sql`
- `apps/cloud/drizzle/0006_panoramic_mother_askani.sql`

No changes to the DBSchema contract in `@executor/sdk` — that already declared these tables in #396. This PR just wires physical Drizzle tables that match.

## Test plan

[...]

PR #401, opened 2026-04-24
feat(api): /executions list, get, tool-calls endpoints

aryasaatvik:
## Stack

**Depends on #400 → #399 → #398 → #396 — merge those first.** Cross-fork GitHub PRs can't use a branch on a contributor fork as a base, so these five PRs display as independent in the UI. The real dependency chain is the commit graph.

| # | PR | Purpose |
|---|---|---|
| 2 | #398 | `feat(execution)`: persist engine runs + tool calls |
| 3 | #399 | `feat(execution)`: trigger propagation (CLI/HTTP/MCP) |
| 5 | **#401 ← you are here** | `feat(api)`: /executions list/get/tool-calls endpoints |

Until the earlier four land, this diff includes their commits. After they merge, this diff shrinks to just the API group + handler additions.

---

## Summary

Extends the existing `/executions` group with the three read endpoints the runs UI needs. Handlers delegate to `executor.executions.*` (added in #396 / #398) and scope each read to the innermost executor scope — same rule the engine applies when writing.

### `GET /executions` — list with filter + cursor + optional meta

Query params:
- `limit` — number, clamped to 1-100 by the SDK.
- `cursor` — opaque base64url token for paginating past the first page.
- `status` — CSV of `pending | running | waiting_for_interaction | completed | failed | cancelled`.
- `trigger` — CSV of trigger kinds. `"unknown"` matches rows with null.
- `tool` — CSV of tool paths or globs (`github.*`).
- `from` / `to` — epoch-ms time bounds.
- `after` — cursor-like "rows after this" filter (used by live mode).
- `code` — case-insensitive substring search.
- `sort` — `<field>,<direction>` (fields: `createdAt`, `durationMs`; directions: `asc`, `desc`).
- `elicitation` — `"true"` / `"false"` to filter to runs that did or didn't elicit.

Meta (facets + timeline buckets + counts) is included **only when the request isn't paginated** (`cursor` + `after` both absent). That makes "first page, full facets" the default call shape and keeps paginated fetches cheap.

### `GET /executions/:id`

Single execution detail + `pendingInteraction`. 404 via `ExecutionNotFoundError` on unknown id.

### `GET /executions/:id/tool-calls`

Tool-call timeline. 404 on unknown execution — guard rail so empty arrays don't mask typos in the path.

## Implementation notes

- Every `Date` serialized to epoch ms at the handler edge (`.getTime()`) so the wire format stays numeric. Schemas mirror the SDK row projections one-to-one modulo that transform.
- CSV / enum parsers (`splitCsv`, `parseSortParam`, `parseElicitationParam`) live in the handler file because they're edge concerns — the SDK takes typed arrays and enums. Invalid sort fields / directions drop to defaults (no 400).

## Test plan

- [x] `bun x tsc --noEmit` across sdk / execution / api / mcp — clean.
- [x] `bun x vitest run` — sdk 97/97, execution 15/15, api 4/4, mcp 23/23.

No new handler-level tests: these are thin wrappers over the SDK store which already has round-trip + filter + meta coverage in `packages/core/sdk/src/executions.test.ts`. The CSV/enum parsers are small enough to validate by inspection; happy to add tests if a reviewer wants them.

### DIFF
--- a/apps/cli/src/main.ts
+++ b/apps/cli/src/main.ts
@@ -394,6 +394,7 @@ const executeCode = (input: {
       payload: {
         code: input.code,
       },
+      headers: { "x-executor-trigger": "cli" },
     });
 
     if (response.status === "paused") {
--- a/apps/cloud/drizzle/0006_panoramic_mother_askani.sql
+++ b/apps/cloud/drizzle/0006_panoramic_mother_askani.sql
@@ -1,0 +1,54 @@
+CREATE TABLE "execution" (
+	"id" text NOT NULL,
+	"scope_id" text NOT NULL,
+	"status" text NOT NULL,
+	"code" text NOT NULL,
+	"result_json" text,
+	"error_text" text,
+	"logs_json" text,
+	"started_at" bigint,
+	"completed_at" bigint,
+	"trigger_kind" text,
+	"trigger_meta_json" text,
+	"tool_call_count" bigint DEFAULT 0 NOT NULL,
+	"created_at" timestamp NOT NULL,
+	"updated_at" timestamp NOT NULL,
+	CONSTRAINT "execution_scope_id_id_pk" PRIMARY KEY("scope_id","id")
+);
+--> statement-breakpoint
+CREATE TABLE "execution_interaction" (
+	"id" text PRIMARY KEY NOT NULL,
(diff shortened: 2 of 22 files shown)


aryasaatvik:

PR #460, opened 2026-05-01
[codex] Use executor-js workspace scope

RhysSullivan:
## Summary

Rename the workspace packages and imports from the old internal `@executor/*` scope to the public `@executor-js/*` scope.

This removes the internal/public double-publish naming split for library packages. The package publisher now publishes workspace package names directly and only rewrites public `workspace:*` dependencies to concrete versions while packing.

## Validation

- `bun install`
- `bun run typecheck`
- `bun run release:publish:packages:dry-run`

## Notes

The repo still has two publish paths because they ship different products: the `executor` CLI wrapper package and the `@executor-js/*` library packages.


e7457166-a11y:
https://github.com/

PR #461, opened 2026-05-01
Plugin internals import from @executor-js/sdk/core

RhysSullivan:
## Summary

- The bare `@executor-js/sdk` path resolves to the Promise build, which only re-exports the small end-user surface (`createExecutor`, ID brands, errors, elicitation types).
- Plugin chunks importing `definePlugin`, `defineSchema`, `StorageError`, `SecretBackedValue`, `SecretBackedMap`, `resolveSecretBackedMap`, `ConnectionId`, etc. from the bare path threw `Export named X not found` at module init when installed from the published tarballs.
- Switch every plugin-internal import to `@executor-js/sdk/core` (the full Effect surface). End-user READMEs and examples keep the bare path.

## Test plan

- [x] Typecheck passes (33/33 packages)
- [x] All seven publishable plugin dists rebuilt; every chunk now references only `@executor-js/sdk/core`
- [ ] CI green

### DIFF
--- a/packages/core/execution/src/description.ts
+++ b/packages/core/execution/src/description.ts
@@ -1,5 +1,5 @@
 import { Effect } from "effect";
-import type { Executor, Source } from "@executor-js/sdk";
+import type { Executor, Source } from "@executor-js/sdk/core";
 
 /**
  * Builds a tool description dynamically.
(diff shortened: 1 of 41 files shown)


RhysSullivan:
Folding into #456 instead — that PR predates the rename and already has this fix (under the old @executor scope). Will rebase #456 onto the post-rename main and re-apply the still-needed commits there.

PR #512, opened 2026-05-04
Visible credential target selector for secret writes; SDK Result import cleanup

RhysSullivan:
Adds `CredentialTargetSelector` to the secret-create dialog, exposing the
full URL-context stack with the labels the plan in
`notes/cloud-workspaces-and-global-sources-plan.md` calls for:

  - Only me in this workspace        → user-workspace
  - Everyone in <workspace name>     → workspace
  - Only me across this org          → user-org
  - Everyone in <org name>           → org

Default selection is the URL context's active write scope (workspace in
workspace context, org in global). Local CLI hosts run a single-scope
stack and the selector renders one disabled option labeled with that
scope.

Server side, `assertScopedWrite` already rejects writes whose scope_id
is outside the URL-resolved stack as a typed `StorageError`; the new
`credential-target.node.test.ts` exercises both halves of the contract:

  - All four legal targets in workspace context succeed and list back
    tagged with the correct scope id.
  - A cross-org write from workspace context is rejected.

Connection writes in v1 still default to the source-add form's
configured `tokenScope` and `bindingScope` (per-user); a future change
can layer this selector into the OAuth setup flow without changing the
server contract.

PR #511, opened 2026-05-04
Reject source writes to personal scopes; visible target selector in add-source forms

RhysSullivan:
Source-definition writes can only target shareable scopes (org or
workspace). Personal scopes (`user_org_*` / `user_workspace_*`) are
reserved for credentials, connections, and policies in the v1 product
model. The SDK now raises `InvalidSourceWriteTargetError` when
`ctx.core.sources.register` is called with a personal scope; the
openapi / mcp / graphql / google-discovery API groups expose this as a
422 on their addSource/addSpec endpoints so clients see a typed
recoverable error rather than a 500.

UI: every add-source form now mounts a `SourceTargetSelector` shared
component (`packages/react/src/plugins/source-target-selector.tsx`) that
renders Workspace / Global options and passes the selected scope id
explicitly to the underlying API call. The selector defaults to the
URL context's active write scope and skips personal scopes, so the
caller never invents a default that the SDK would reject.

The cloud test harness covers the legal write paths (workspace and org
targets from workspace context). The personal-scope rejection has SDK
test coverage in `executor.test.ts`; the HTTP layer wires the same
error through with `httpApiStatus: 422`.

PR #510, opened 2026-05-04
Surface overridden sources in sources.list, split sidebar by scope

RhysSullivan:
The cloud workspace sidebar now splits into a Workspace section and a
Global section, with global sources that a workspace shadows rendered as
muted entries with an `Overridden` badge.

To make this possible, `executor.sources.list()` no longer drops shadowed
outer rows. Both the inner (effective) and outer (shadowed) rows ship
back, and the shadowed row carries `overriddenBy: <innerScopeId>` so the
UI can render the Overridden state without re-deriving the precedence
rule. Tool invocation continues to pick the innermost row (covered by
the existing `tools.invoke picks the innermost tool ...` test).

The same split lands on the standalone sources page. Local hosts run a
single-scope stack so the bucket render is gated on the presence of a
`workspace_*` entry in the scope stack.

Adds a node test that creates a workspace source shadowing a global one
and asserts both rows are returned with correct scopeId + overriddenBy
fields.

PR #509, opened 2026-05-04
Add activeWriteScopeId to scope.info, expose useActiveWriteScopeId

RhysSullivan:
The scope handler now returns three things: the active display scope
(`id`/`name`/`dir`), a pre-computed `activeWriteScopeId` (the default
source-definition write target — `org` in global, `workspace` in workspace
contexts), and the full innermost-first stack for storage-target selectors.

The active write scope is computed by skipping personal scopes (those whose
id is prefixed `user_org_` or `user_workspace_`) and picking the first
non-personal scope from the inner end. That collapses to `org` in the
global stack `[user_org, org]` and to `workspace` in the workspace stack
`[user_workspace, workspace, user_org, org]`.

Client side, expose `useActiveWriteScopeId()` for default writes; keep
`useScope()` as an alias so plugin and existing pages don't churn. Update
the cloud shell + every page in `packages/react/src/pages/` to call the
explicit hook so intent is clear.

PR #513, opened 2026-05-04
Policies expose the full storage stack with target selector

RhysSullivan:
Policies expose the full storage stack with target selector

The SDK's `executor.policies.create` and `policies.list` already accept
every scope in the URL-resolved stack (`assertScopedWrite` validates
membership; `resolveToolPolicy` ranks by `scopeRank` so the innermost
matching row wins). What was missing was the visible UI surface — the
add-policy form pinned writes to the active write scope.

`AddPolicyForm` now renders a `CredentialTargetSelector` ("Apply to") so
users can target any of the four levels in workspace context (or both in
global). The page-level `scopeId` continues to drive the optimistic
list/cache family; the form passes the explicit target through to the
API call.

`policies-stack.node.test.ts` pins both invariants:

  - All four scope levels accept policy writes from workspace context,
    and listing tags each row with its owning scope.
  - When the same pattern is written at multiple scopes, `policies.list`
    returns rows sorted innermost-first — the executor's invocation
    path consumes that ordering for "innermost wins" precedence.

Build HttpApiBuilder.layer per protected mount so workspace prefix registers

Effect's Layer system memoizes one ProtectedCloudApiLive instance, so
piping it through both the /:org and /:org/:workspace prefixed router
layers only registered the first prefix's routes — workspace requests
fell through to RouteNotFound. Each mount now constructs its own
HttpApiBuilder.layer(ProtectedCloudApi) so both prefixes install.

Also extends the McpExtension surface and the matching api/group entry
for refreshSource to allow InvalidSourceWriteTargetError, mirroring the
addSource failure union added in chunk 11.

Mount OrgHttpApi at workspace prefix too

The URL-context fetch wrapper in the react package always prefixes
outgoing /api/... requests with the page's full URL handle pair, so
from a /:org/:workspace page even org-level endpoints (workspaces list,
org members) come through with the workspace prefix. Mount OrgHttpApi
at both /:org and /:org/:workspace so those requests resolve.

Each mount builds its own HttpApiBuilder.layer instance — Effect's
Layer system memoizes one instance otherwise and only the first prefix
registers routes (same trick we needed for the protected API).

Resolve active workspace from URL params for callers above the layout

Shell + UserFooter live in the /$org layout's render tree, ABOVE the
WorkspaceRouteProvider rendered by /$org/$workspace. So the trigger
label always saw a null context and showed "Global" even when the URL
had a workspace segment.

useOptionalWorkspaceRoute now falls back to URL params + workspacesAtom
when the React context isn't set, so any caller rendered under /$org
sees the active workspace as long as the URL has the slug. Direct
context lookup is preferred when the layout has already resolved the
workspace, to skip the atom roundtrip.

Prefix shared-package routes with active URL context via useAppHref

The shared @executor-js/react pages and the command palette hard-coded
absolute Link paths (/sources/add/$pluginKey, /sources/$namespace,
/policies) against the local app's flat route tree. On cloud those
routes live under /:org or /:org/:workspace, so clicking landed
nowhere matching and routes/index.tsx redirected back to the user's
first org.

useAppHref(path, params?, search?) reads the active prefix from
useLocation().pathname and returns a fully-qualified href, falling back
to the unmodified path on local. command-palette, sources page (3
links + the navigate call), and tools page now route through it.

Reserve org sub-route names so /:org/sources isn't read as a workspace

The fetch wrapper and href helpers split window.location.pathname into
:org / :workspace assuming the second segment is a workspace slug. On
/rhys-org/sources/add/openapi that turned 'sources' into a workspace,
so client requests for /api/scope ended up at /api/rhys-org/sources/scope
and 404'd.

Adds RESERVED_SECOND_SEGMENTS (sources, connections, secrets, policies,
tools, plus the org-admin marker '-') in a shared parseUrlContext.
Workspace slugs can't collide with these reserved names — same
constraint TanStack's file-routes already imposes — and both the fetch
prefixer and the new AppLink resolve URL context the same way.

Also introduces <AppLink> as a TanStack Link wrapper that prepends the
context prefix to absolute 'to' strings. The cloud-shared package now
uses AppLink instead of hand-spliced strings via useAppHref for static
hrefs (sources detail, source-add presets, manage-policies button).

Make ContextAwareHttpClient the single source of truth for /api prefixing

Plugin clients (openapi, mcp, graphql, onepassword, google-discovery)
all snapshot getBaseUrl() at module load. routes/$org.tsx + routes/
$org/$workspace.tsx then mutated the same module variable via
setBaseUrl, racing module load order — when a plugin module loaded
AFTER setBaseUrl fired it baked /api/${handle} into outgoing URLs.
The fetch wrapper then prefixed AGAIN, producing
/api/${handle}/${workspace}/${handle}/scopes/... and 404s.

createPluginAtomClient now accepts an httpClient option. Each plugin
client passes ContextAwareHttpClient so URL-rewriting is consistent
across the executor + cloud auth + plugin clients. Drops setBaseUrl
entirely; getBaseUrl is back to a pure function returning
${origin}/api. Removes the setBaseUrl calls from /$org.tsx and
/$org/$workspace.tsx — they're no longer needed since the wrapper
handles per-request prefixing.

Refresh scopeAtom when URL context changes

scopeAtom caches /scope responses for 5 minutes. The cloud middleware
returns different scope stacks for /api/:org/scope (global) vs
/api/:org/:workspace/scope (workspace), so the cached value is stale
the moment the user navigates between contexts — the SourceTargetSelector
on /:org/:workspace/sources/add/openapi only saw the org stack and
showed Global as the only option.

ScopeProvider now reads useLocation(), derives a context key
(org-handle / workspace-slug pair), and triggers useAtomRefresh on
change. Stays cached within a single context — same-context page
navigation doesn't refetch.

### DIFF
--- a/apps/cloud/src/api/layers.ts
+++ b/apps/cloud/src/api/layers.ts
@@ -64,25 +64,50 @@ export const makeNonProtectedApiLive = (
 // Routes scoped to a specific org (membership management, switching, etc.).
 // Auth is enforced by `OrgAuth` middleware declared on `OrgHttpApi`.
 //
-// OrgHttpApi mounts under `/api/:org/...` so workspace endpoints are
-// addressable per-org (`POST /api/:org/workspaces`,
-// `GET /api/:org/workspaces/:slug`). `start.ts` strips the leading `/api`
-// before forwarding, so the prefix here is `/:org` (not `/api/:org`).
+// OrgHttpApi mounts under `/api/:org/...` AND `/api/:org/:workspace/...` so
+// the same org-level endpoints (workspaces list/create, members, etc.) stay
+// reachable from either context — the URL-context fetch wrapper in the
+// react package always prefixes outgoing `/api/...` requests with the
+// page's URL handle pair. `start.ts` strips the leading `/api` before
+// forwarding, so the prefixes here omit it.
+//
+// Each mount needs its own `HttpApiBuilder.layer(OrgHttpApi)` instance
+// because Effect's Layer system memoizes a shared instance and only the
+// first prefix's routes would register otherwise (see
+// `apps/cloud/src/api/protected.ts` for the same pattern).
 const OrgPrefixedRouterLayer = Layer.effect(HttpRouter.HttpRouter)(
   Effect.map(HttpRouter.HttpRouter.asEffect(), (router) =>
     router.prefixed("/:org"),
   ),
 );
 
-export const makeOrgApiLive = (
-  rsLive: Layer.Layer<DbService | UserStoreService>,
-) =>
+const OrgWorkspacePrefixedRouterLayer = Layer.effect(HttpRouter.HttpRouter)(
+  Effect.map(HttpRouter.HttpRouter.asEffect(), (router) =>
+    router.prefixed("/:org/:workspace"),
+  ),
+);
+
+const makeOrgLayer = () =>
   HttpApiBuilder.layer(OrgHttpApi).pipe(
     Layer.provide(Layer.mergeAll(OrgHandlers, WorkspacesHandlers)),
-    Layer.provide(requestScopedMiddleware(rsLive).layer),
+  );
+
+export const makeOrgApiLive = (
+  rsLive: Layer.Layer<DbService | UserStoreService>,
+) => {
+  const requestScopedLayer = requestScopedMiddleware(rsLive).layer;
+  const orgMount = makeOrgLayer().pipe(
+    Layer.provide(requestScopedLayer),
     Layer.provideMerge(OrgAuthLive),
     Layer.provide(OrgPrefixedRouterLayer),
   );
+  const workspaceMount = makeOrgLayer().pipe(
+    Layer.provide(requestScopedLayer),
+    Layer.provideMerge(OrgAuthLive),
+    Layer.provide(OrgWorkspacePrefixedRouterLayer),
+  );
+  return Layer.mergeAll(orgMount, workspaceMount);
+};
 
 // Default exports use the production per-request layer. Existing callers
 // that import `NonProtectedApiLive`/`OrgApiLive` continue to work; the
--- a/apps/cloud/src/api/protected.ts
+++ b/apps/cloud/src/api/protected.ts
@@ -2,12 +2,16 @@
 // because `makeExecutionStack` imports `cloudflare:workers`, which the test
 // harness can't load in the workerd test runtime.
 
-import { HttpApiSwagger } from "effect/unstable/httpapi";
+import { HttpApiBuilder, HttpApiSwagger } from "effect/unstable/httpapi";
 import {
   HttpRouter,
   HttpServerRequest,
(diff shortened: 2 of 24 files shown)


RhysSullivan:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/513?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#513** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/513?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/513?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#512** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/512?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#511** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/511?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#510** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/510?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#509** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/509?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#508** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/508?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#507** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/507?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#514](https://github.com/RhysSullivan/executor/pull/514) <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/514?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#506** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/506?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#505** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/505?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#504** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/504?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#503** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/503?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#502** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/502?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #538, opened 2026-05-05
[codex] Preserve storage memory adapter failures

RhysSullivan:
## Summary
- route malformed memory adapter `in` / `not_in` predicates through `StorageError` instead of throwing
- keep filtering in the Effect error channel so invalid query state fails explicitly
- preserve existing match behavior for valid predicates

## Validation
- `bun run typecheck` in `packages/core/storage-core`
- `bunx vitest run` in `packages/core/storage-core`
- `bunx oxlint -c .oxlintrc.jsonc packages/core/storage-core/src/testing/memory.ts --deny-warnings`

Stacked on #535.

### DIFF
--- a/packages/core/storage-core/src/testing/memory.ts
+++ b/packages/core/storage-core/src/testing/memory.ts
@@ -25,17 +25,14 @@ import type {
   StorageFailure,
 } from "../adapter";
 import type { DBSchema } from "../schema";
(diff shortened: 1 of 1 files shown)


RhysSullivan:
Closing as superseded by #646, which covers the same lint file set in the newer audited cleanup batch.

PR #549, opened 2026-05-06
Parse GraphQL store rows with schema

RhysSullivan:
## Summary
- replace JSON.parse for persisted operation bindings with Effect Schema parsing
- decode GraphQL source, operation, and child rows at the store boundary
- remove redundant primitive casts from usage and source lookup paths

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/plugins/graphql/src/sdk/store.ts --format json
- bun run typecheck (packages/plugins/graphql)
- bunx vitest run src/sdk/plugin.test.ts

### DIFF
--- a/packages/plugins/graphql/src/sdk/store.ts
+++ b/packages/plugins/graphql/src/sdk/store.ts
@@ -1,4 +1,4 @@
-import { Effect } from "effect";
+import { Effect, Schema } from "effect";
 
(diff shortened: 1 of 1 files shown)


RhysSullivan:
Closing as superseded by #641, which covers the same lint file set in the newer audited cleanup batch.

PR #555, opened 2026-05-06
Clean up Slack contact route boundaries

RhysSullivan:
## Summary
- parse Slack contact request bodies through Effect Schema before field handling
- yield HttpResponseError values directly from the Effect generator
- remove unknown error stringification from Turnstile/rate-limit fallbacks and server-error logging

## Verification
- bunx oxlint -c .oxlintrc.jsonc apps/cloud/src/api/slack.ts --format json
- bun run typecheck (apps/cloud)
- bunx vitest run --config vitest.node.config.ts src/services/sources-api.node.test.ts


RhysSullivan:
Closing as superseded by #666, which covers the same lint file set in the newer audited cleanup batch.

PR #558, opened 2026-05-06
Clean up Cloud API error response mapping

RhysSullivan:
## Summary
- replace HttpResponseError instanceof mapping with a tagged guard
- avoid instanceof Error when logging server response failures
- keep default internal error mapping stable for unknown failures

## Verification
- bunx oxlint -c .oxlintrc.jsonc apps/cloud/src/api/error-response.ts --format json
- bun run typecheck (apps/cloud)
- bunx vitest run --config vitest.node.config.ts src/services/sources-api.node.test.ts

### DIFF
--- a/apps/cloud/src/api/error-response.ts
+++ b/apps/cloud/src/api/error-response.ts
@@ -1,8 +1,5 @@
-import { Cause, Data, Effect, Result } from "effect";
-import {
-  HttpServerRespondable,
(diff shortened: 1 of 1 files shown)


RhysSullivan:
Closing as superseded by #666, which covers the same lint file set in the newer audited cleanup batch.

PR #562, opened 2026-05-06
Clean up Google Discovery invocation failures

RhysSullivan:
## Summary
- return typed invocation failures for missing path parameters instead of throwing
- remove unknown/native error text from user-facing HTTP/body/OAuth messages
- preserve original causes on GoogleDiscoveryInvocationError where the type supports it
- yield tagged invocation errors directly from the generator

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/plugins/google-discovery/src/sdk/invoke.ts --format json
- bun run typecheck (packages/plugins/google-discovery)
- bunx vitest run src/sdk/plugin.test.ts src/sdk/document.test.ts src/api/handlers.test.ts


RhysSullivan:
Closing as superseded by #648, which covers the same lint file set in the newer audited cleanup batch.

PR #566, opened 2026-05-06
Clean up OAuth handler error mapping

RhysSullivan:
## Summary
- replace OAuth popup error instanceof checks with schema-backed decoding
- keep secret-backed config resolution failures typed before resolving maps
- return OAuthCompleteError from callback failure mapping instead of new Error

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/core/api/src/handlers/oauth.ts --format json
- bun run typecheck (packages/core/api)
- bunx vitest run src/oauth-popup.test.ts


RhysSullivan:
Closing as superseded by #590, which covers the same lint file set in the newer audited cleanup batch.

PR #568, opened 2026-05-06
Clean up storage-core factory failures

RhysSullivan:
## Summary
- replace storage factory sync throws and generator Effect.fail(new ...) with typed StorageError yields
- decode JSON-backed values through Effect Schema instead of JSON.parse
- keep transform failure messages stable while preserving raw causes internally
- add conformance coverage that unknown models fail before backend dispatch

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/core/storage-core/src/factory.ts packages/core/storage-core/src/testing/conformance.ts --format json
- bun run typecheck (packages/core/storage-core)
- bunx vitest run src/memory.test.ts

### DIFF
--- a/packages/core/storage-core/src/factory.ts
+++ b/packages/core/storage-core/src/factory.ts
@@ -25,7 +25,7 @@
 //     to running the callback against the current adapter
 // ---------------------------------------------------------------------------
 
-import { Effect } from "effect";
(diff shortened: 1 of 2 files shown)


RhysSullivan:
Closing as superseded by newer audited cleanup PRs covering the same lint file set: #638, #646.

PR #569, opened 2026-05-06
Clean up storage-drizzle typed failures

RhysSullivan:
## Summary
- convert drizzle table and where compilation failures into typed StorageError values
- keep driver error messages stable without instanceof Error or String(cause)
- preserve transient retry classification for string driver causes
- keep the native drizzle rollback throw as a narrow boundary exception

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/core/storage-drizzle/src/adapter.ts --format json
- bun run typecheck (packages/core/storage-drizzle)
- bunx vitest run src/adapter.test.ts

### DIFF
--- a/packages/core/storage-drizzle/src/adapter.ts
+++ b/packages/core/storage-drizzle/src/adapter.ts
@@ -11,7 +11,7 @@
 //     generation, encode/decode all happen in storage-core
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


RhysSullivan:
Closing as superseded by #638, which covers the same lint file set in the newer audited cleanup batch.

PR #573, opened 2026-05-06
Clean up MCP host typed boundaries

RhysSullivan:
## Summary
- replace MCP host manual tagged checks with Predicate/Match helpers
- wrap elicitInput promise failures in a typed boundary error while preserving cancel fallback
- parse resume JSON content through Effect Schema
- keep debug logging try/catch as a narrow callback boundary

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/hosts/mcp/src/server.ts --format json
- bun run typecheck (packages/hosts/mcp)
- bunx vitest run src/server.test.ts


RhysSullivan:
Closing as superseded by #659, which covers the same lint file set in the newer audited cleanup batch.

PR #574, opened 2026-05-06
Clean up Google Discovery plugin typed failures

RhysSullivan:
## Summary
- replace Google Discovery plugin tagged-error instanceof checks with Predicate handling
- convert discovery fetch failures to typed GoogleDiscoverySourceError values with stable messages
- remove URL parsing try/catch and redundant scope casts
- derive the extension type from the local extension factory for this file’s active lint

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/plugins/google-discovery/src/sdk/plugin.ts --format json
- bun run typecheck (packages/plugins/google-discovery)
- bunx vitest run src/sdk/plugin.test.ts

### DIFF
--- a/packages/plugins/google-discovery/src/sdk/plugin.ts
+++ b/packages/plugins/google-discovery/src/sdk/plugin.ts
@@ -1,4 +1,4 @@
-import { Effect, Option } from "effect";
+import { Effect, Option, Predicate, Schema } from "effect";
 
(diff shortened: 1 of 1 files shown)


RhysSullivan:
Closing as superseded by #657, which covers the same lint file set in the newer audited cleanup batch.

PR #594, opened 2026-05-06
Parse plugin loader config with schema

RhysSullivan:
## Summary
- parse plugin loader JSONC output through the existing Effect Schema config model
- keep loader rejection messages stable at the Promise boundary
- retain the require.resolve adapter as an explicit narrow boundary

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/core/config/src/load-plugins.ts --deny-warnings
- git diff --check
- bun run typecheck (packages/core/config)
- bunx --bun vitest run src/load-plugins.test.ts (packages/core/config)


RhysSullivan:
Closing as superseded by #610, which covers the same lint file set in the newer audited cleanup batch.

PR #597, opened 2026-05-06
Use typed failures in memory adapter filters

RhysSullivan:
## Summary
- convert invalid memory adapter in/not_in filter values into StorageError failures
- thread filter evaluation through Effect so failures stay in the storage error channel
- preserve AND/OR short-circuit behavior while removing raw throw/new Error

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/core/storage-core/src/testing/memory.ts --deny-warnings
- git diff --check
- bun run typecheck (packages/core/storage-core)
- bunx --bun vitest run src/memory.test.ts (packages/core/storage-core)

### DIFF
--- a/packages/core/storage-core/src/testing/memory.ts
+++ b/packages/core/storage-core/src/testing/memory.ts
@@ -26,6 +26,7 @@ import type {
 } from "../adapter";
 import type { DBSchema } from "../schema";
 import { createAdapter } from "../factory";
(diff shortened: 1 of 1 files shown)


RhysSullivan:
Closing as superseded by #646, which covers the same lint file set in the newer audited cleanup batch.

PR #608, opened 2026-05-06
Parse local config sync with schema

RhysSullivan:
## Summary
- decode executor.jsonc source replay input with Effect Schema
- remove redundant scope id casting
- log config sync failure causes without stringifying unknown errors

## Verification
- bunx oxlint -c .oxlintrc.jsonc apps/local/src/server/config-sync.ts --deny-warnings
- git diff --check
- bun run typecheck (apps/local)


RhysSullivan:
Closing as superseded by #620, which covers the same lint file set in the newer audited cleanup batch.

PR #616, opened 2026-05-06
Parse connection migration JSON with Schema

RhysSullivan:
## Summary
- replace migration JSON.parse try/catch paths with Effect Schema JSON decoders
- wrap SQLite transaction failures into typed migration errors while keeping rollback throws at the boundary

## Verification
- bunx oxlint -c .oxlintrc.jsonc apps/local/src/server/migrate-connections.ts --deny-warnings
- bun run --cwd apps/local typecheck
- bunx --bun vitest run src/server/migrate-connections.test.ts


RhysSullivan:
Closing as superseded by #659, which covers the same lint file set in the newer audited cleanup batch.

PR #640, opened 2026-05-06
Parse OpenAPI store rows with Schema

RhysSullivan:
## Summary
- replace JSON.parse and primitive row casts in the OpenAPI store with Effect Schema decoders
- preserve legacy header and OAuth normalization behavior
- yield tagged storage errors directly from Effect generators

## Verification
- bunx oxlint -c .oxlintrc.jsonc packages/plugins/openapi/src/sdk/store.ts --deny-warnings
- bunx oxfmt --check packages/plugins/openapi/src/sdk/store.ts
- bun run --cwd packages/plugins/openapi typecheck
- bun run --cwd packages/plugins/openapi test

### DIFF
--- a/packages/plugins/openapi/src/sdk/store.ts
+++ b/packages/plugins/openapi/src/sdk/store.ts
@@ -1,4 +1,4 @@
-import { Effect, Schema } from "effect";
+import { Effect, Option, Schema } from "effect";
 
 import {
   ConnectionId,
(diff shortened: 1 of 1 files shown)


RhysSullivan:
Closing as superseded by #656, which covers the same lint file set in the newer audited cleanup batch.

PR #715, opened 2026-05-08
Bind execution continuations to callers

RhysSullivan:
## Summary
- add optional owner ids and lifetime limits for paused executions
- pass continuation ownership through the MCP host and cloud MCP session setup
- keep the model-visible resume tool disabled by default while preserving explicit host configuration

## Verification
- cd packages/core/execution && bunx vitest run src/tool-invoker.test.ts
- cd packages/hosts/mcp && bunx vitest run src/server.test.ts
- cd apps/cloud && bunx vitest run src/mcp-flow.test.ts
- bun run format:check
- bun run lint
- bun run typecheck
- bun run test


RhysSullivan:
Closing this from the active stack. The continuation/resume behavior is intentionally accepted for now because the cloud MCP path runs inside Durable Object session ownership, with resume tool exposure also acceptable for the current product model.

PR #761, opened 2026-05-11
Fix crash on variadic tool path arguments

### LINKED ISSUE #738: `executor call` crashes because variadic path args do not have `.at`
I’m seeing `executor call` crash before it invokes the tool.

Version: `executor@1.4.15`

Command:

```sh
executor call paperless profile profileRetrieve
```

This also fails when including a JSON arg:

```sh
executor call paperless profile profileRetrieve '{}'
```

Error:

```text
e.rawPathParts.at is not a function. (In 'e.rawPathParts.at(-1)', 'e.rawPathParts.at' is undefined)
```

This looks like it comes from `resolveToolInvocation` here:

https://github.com/RhysSullivan/executor/blob/f1b74b7/apps/cli/src/main.ts#L1099

```ts
const maybeJsonArg = input.rawPathParts.at(-1)?.trim();
```

`rawPathParts` is typed as `ReadonlyArray<string>`, but at runtime it does not seem to be a real array in the installed CLI. My guess is that `Args.variadic(Args.string("tool-path-segment"))` is returning an array-like or Effect collection in this runtime, so `.at` is missing.

This makes `executor call` unusable, even for a no-argument tool path.

grfwings:
Fixes #738

Effect CLI currently requires an explicit options object for `Argument.variadic`; without it, the parsed `pathParts` value
 becomes `{}` instead of an array. That made `executor call ...` crash before invoking the tool.

 This updates the call command to parse path segments with:
 ```ts
   Args.string("tool-path-segment").pipe(Args.variadic({}))
 ```
 and adds an explicit failure if the parser ever returns a non-array value.
 
 Upstream issue: Effect-TS/effect#6228

Also, added `.pi/` to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ MISTAKES.md
 DESIRES.md
 LEARNINGS.md
 .alchemy/
+
+# Pi coding agent
+.pi/
--- a/apps/cli/src/main.ts
+++ b/apps/cli/src/main.ts
@@ -1115,6 +1115,12 @@ const resolveToolInvocation = (input: {
   rawPathParts: ReadonlyArray<string>;
 }): Effect.Effect<{ path: string; args: Record<string, unknown> }, Error> =>
   Effect.gen(function* () {
+    if (!Array.isArray(input.rawPathParts)) {
+      return yield* Effect.fail(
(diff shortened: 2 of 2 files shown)


RhysSullivan:
thanks!

PR #763, opened 2026-05-11
fix(cli): invoke Args.variadic() so executor call parses positional args

DomiR:
executor call <path...> always crashes with rawPathParts.at is not a function.

Args.variadic is dual function and otherwise needs second (e.g. undefined) option.


DomiR:
You were quicker 💪

PR #764, opened 2026-05-11
mount Toaster in local app

### LINKED ISSUE #760: Secret deletion error doesn't surface
## Description

When attempting to delete a secret that is currently referenced by a source, the API returns an error (because the secret is in use), but the UI doesn't surface this error to the user. The delete button appears to do nothing — no error toast, no message, no visual feedback.

## Steps to Reproduce

1. Create a secret (e.g., a GitHub PAT)
2. Attach it to a source (e.g., as an Authorization header)
3. Try to delete the secret from the secrets management UI
4. Nothing visually happens and the secret remains, with no indication of why

## Expected Behavior

The UI should show an error message explaining that the secret cannot be deleted because it is in use by one or more sources, and ideally name which source(s) reference it.

## Actual Behavior

The delete fails silently. The error is only visible in the browser's Network tab.

## Additional Context

Related UX gap: secret *values* also can't be edited after creation. The current workaround is to create a new secret with the updated value and re-attach it to the source, then delete the old one. Being able to edit secret values in place would make this workflow much smoother, but that may be a separate issue.

Oladejiraji:
## Summary

- Toast notifications (e.g. error when removing an in-use secret) were silently swallowed in the local app because the `<Toaster />` component was never mounted in the React tree
- Added `<Toaster />` from `@executor-js/react/components/sonner` to the local app root layout, matching the cloud app

Closes #760

## Test plan

- [ ] Open the local app, create a secret, attach it to a source, then try to remove the secret — verify the "Secret is used by N sources" error toast appears
- [ ] Verify other toast calls (e.g. connection removal errors) also surface correctly

### DIFF
--- a/apps/local/src/routes/__root.tsx
+++ b/apps/local/src/routes/__root.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { createRootRoute } from "@tanstack/react-router";
 import { ExecutorProvider } from "@executor-js/react/api/provider";
 import { ExecutorPluginsProvider } from "@executor-js/sdk/client";
+import { Toaster } from "@executor-js/react/components/sonner";
 import { plugins as clientPlugins } from "virtual:executor/plugins-client";
 import { Shell } from "../web/shell";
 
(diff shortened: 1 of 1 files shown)


RhysSullivan:
thanks!

PR #766, opened 2026-05-11
fix(openapi): wrap wire-shape OAuth2 config before storage encode

RhysSullivan:
## Summary

- `addSpec`/`updateSource` crashed whenever the request carried an `oauth2` block, surfacing to users as an opaque `{"_tag":"InternalError","traceId":"…"}` after OAuth.
- Root cause: the HTTP edge validates the OAuth2 payload against the Struct shape of `OAuth2SourceConfigSchema`, so the plugin received a plain object. Storage encodes via the Class schema, whose `instanceof` check rejected it with `Expected OpenApiOAuth2SourceConfig, got {…}`. The Postgres txn aborted as `StorageError: pg transaction failed`, observability translated to `InternalError({ traceId })`.
- Fix: `canonicalizeOAuth2` is the single funnel both `addSpec` (via `rebuildSource`) and `updateSource` go through, so wrapping any non-instance input as `new OAuth2SourceConfig(...)` there fixes both callers without touching the public API surface. SDK callers that already pass a class instance short-circuit through the instanceof check.
- Reproduced in Sentry as `NODE-CLOUDFLARE-WORKERS-3N` (Google + Vercel users this week, and reported again today for Microsoft SharePoint). Existing tests didn't catch it because every test constructed `new OAuth2SourceConfig({...})` directly and never exercised the wire path.

## Test plan

- [x] New regression test in `plugin.test.ts` simulates the API boundary by handing `addSpec` a plain object that satisfies the Struct shape, then asserts the source round-trips back through `getSource`. On `main` it fails with the exact production message at `store.ts:553`; with the fix it passes.
- [x] `bun run vitest run` for the openapi plugin (93 tests pass).
- [x] `bun run lint`, `bun run format:check`, `bun run typecheck` clean.

### DIFF
--- a/packages/plugins/openapi/src/sdk/plugin.test.ts
+++ b/packages/plugins/openapi/src/sdk/plugin.test.ts
@@ -523,6 +523,59 @@ layer(TestLayer)("OpenAPI Plugin", (it) => {
     }),
   );
 
+  // Regression: the HTTP edge validates `oauth2` against the Struct shape of
+  // `OAuth2SourceConfigSchema` and hands the plugin a plain object — not an
+  // `OAuth2SourceConfig` instance. Storage encodes via the Class schema, which
+  // does an `instanceof` check, so without `canonicalizeOAuth2` wrapping the
+  // input the request crashed with "Expected OpenApiOAuth2SourceConfig, got
+  // {...}", the txn rolled back as `StorageError: pg transaction failed`, and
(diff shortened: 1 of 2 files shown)


RhysSullivan:
Closing in favor of the wholesale Schema.Class ban — the underlying foot-gun (Effect 4.x instanceof check + TypeScript structural typing) needs to be removed at the source, not patched per-site.

PR #774, opened 2026-05-12
fix(oauth): surface real callback error in popup with collapsible details

### LINKED ISSUE #773: [UX] OAuth callback errors are swallowed; popup always shows generic "Authentication failed"
#### Summary

The OAuth callback handler replaces every completion error with hardcoded `"Authentication failed"` before rendering the popup. The real cause is logged server-side but never reaches the user, making any OAuth failure (provider quirks, token exchange, DCR, scope mismatches) undiagnosable from the UI. Provider-agnostic — affects every MCP server using OAuth.

#### Where

`packages/core/api/src/handlers/oauth.ts:175-182`

```ts
.pipe(
  Effect.tapError((cause) =>
    Effect.logError("OAuth callback completion failed", cause),
  ),
  Effect.catchCause(() =>
    Effect.fail(new OAuthCompleteError({ message: "Authentication failed" })),
  ),
),
```

`catchCause` discards the original cause for a placeholder, so by the time `toPopupErrorMessage` runs there's only the placeholder left to decode.

#### Repro

Any provider whose token endpoint returns a non-200 status, unexpected body shape, or otherwise trips `oauth4webapi`'s `processAuthorizationCodeResponse` validators.

Real example: Supabase's `/v1/oauth/token` returns `HTTP 201 Created` instead of the spec-required `200 OK` (upstream fix scheduled 2026-05-26). `oauth4webapi` rejects with `UnexpectedResponseStatus`, the error is wrapped as `OAuthCompleteError("Token exchange failed: ...")` at `packages/core/sdk/src/oauth-service.ts:911`, then blanked at the callback handler.

Server log (with the actual cause):

```
ERROR: OAuth callback completion failed
  message: Token exchange failed: OAuth token exchange failed: "response" is not a conform Token Endpoint response (unexpected HTTP status code) (HTTP 201 Created from https://api.supabase.com/v1/oauth/token; ...)
  _tag: "OAuthCompleteError"
```

UI: `Authentication failed`.

#### Why it's worth fixing

- Users can't distinguish "non-compliant provider", "wrong redirect URI", "bad client_secret", "expired session", or "network blip".
- Local installs surface the real cause in terminal logs, but it's buried.
- Cloud/hosted deployments have no server logs visible to end users — the popup is the only diagnostic surface, and executor is multi-user, so the affected user needs enough info to act.
- The decoding pipeline (`toPopupErrorMessage` → `decodeOAuthCompleteError`) already surfaces `OAuthCompleteError.message`; `catchCause` is the only thing in the way. Start/probe paths already propagate the real error fine — this is specific to the popup-callback codepath.

#### Proposed fix

1. **Stop swallowing the cause**: drop the `Effect.catchCause` wrapper in `packages/core/api/src/handlers/oauth.ts` so the original tagged error (`OAuthCompleteError` / `OAuthSessionNotFoundError`) propagates with its real `.message`.
2. **Two-tier popup**: render a short, tag-derived headline (e.g. `OAuthCompleteError → "Could not complete authentication"`) with the full technical message inside a collapsed `<details>` disclosure. Skip auto-close on failure so users can expand and read.

Raw `oauth4webapi` wording is long and technical; pushing it straight into the headline would regress the other way. Two-tier gives a stable summary plus diagnostic detail.

#### Security note

Surfacing the real error and not leaking credentials are in tension, especially in shared cloud deployments where teammates or admins may see the popup. The existing redaction in `failOAuth2WithHttpSummary` (`packages/core/sdk/src/oauth-helpers.ts:264`) strips `access_token`, `refresh_token`, `id_token`, `client_secret`, and authorization-code values from response bodies before they reach the error, so the popup is safe to render today. Any new error-string source must go through the same redaction.

#### Out of scope

- The Supabase 201 quirk itself — upstream, fix announced for 2026-05-26.
- Other providers' non-compliance — should still fail, but loudly enough that users know what's wrong.
- Pattern-matching specific library error shapes (`UnexpectedResponseStatus`, `invalid_grant`, etc.) into bespoke short messages — the tag-based summary already gives a stable headline; per-shape branches would be a maintenance pit for marginal benefit.

Mark-Life:
Closes #773

## Summary

OAuth callback failures rendered a hardcoded `"Authentication failed"` because the handler replaced the original cause via `Effect.catchCause`, making every integration failure undiagnosable from the UI. This change drops the `catchCause` so the original tagged error propagates, and adds a two-tier popup: a tag-derived headline plus the full technical message inside a collapsed `<details>` disclosure. Auto-close is skipped on failure so users can expand and read.

Popup with technical error

## Why

The decoding pipeline already knew how to surface `OAuthCompleteError.message`; the swallowing was the only thing in the way. Raw `oauth4webapi` wording is too noisy for a headline, hence the two-tier presentation. Backwards-compatible — existing consumers reading `result.error` keep working.

## Changes

- `packages/core/sdk/src/oauth-popup-types.ts` — adds `errorDetails?: string` to the failure variant.
- `packages/core/api/src/oauth-popup.ts` — new `PopupErrorMessage` type, popup HTML renders the disclosure with dark-mode styling and HTML escaping, auto-close skipped on failure.
- `packages/core/api/src/handlers/oauth.ts` — drops `Effect.catchCause`; `toPopupErrorMessage` returns `{ short, details? }` derived from the tag.
- `packages/core/api/src/oauth-popup.test.ts` — covers disclosure rendering, XSS escaping of `errorDetails`, omission when `details === short`, and the new mapper signature.

## Test plan

- [x] `bun run --cwd packages/core/api test` — 23/23 pass.
- [x] `bun run lint` — clean.
- [x] `bun run format:check` — clean.
- [x] `bun run typecheck` on `core/api` and `core/sdk` — no new diagnostics (only pre-existing effect-rule suggestions in unrelated files).
- [ ] Manually trigger an OAuth callback failure (e.g. force a non-200 from a token endpoint) and verify the popup shows the short headline plus an expandable Details block.

### DIFF
--- a/.changeset/oauth-popup-error-details.md
+++ b/.changeset/oauth-popup-error-details.md
@@ -1,0 +1,5 @@
+---
+"executor": patch
+---
+
+Surface the real OAuth callback error in the popup with a collapsible details disclosure. See `apps/cli/release-notes/next.md`.
--- a/apps/cli/release-notes/next.md
+++ b/apps/cli/release-notes/next.md
@@ -11,3 +11,7 @@ Two regressions kept `executor.jsonc` and the runtime DB from agreeing on which 
 ### Variadic tool path arguments no longer crash
 
 Calling a tool with multiple positional path arguments (`executor <tool> path/a path/b ...`) no longer panics in the CLI argument parser. Thanks @grfwings (#761)
+
+### OAuth popup surfaces the real callback error
+
+OAuth callback failures previously rendered a hardcoded `"Authentication failed"`, hiding the actual cause behind a generic placeholder. The popup now shows a short tag-derived headline plus the full technical message inside a collapsible `<details>` disclosure, and skips auto-close on failure so users can read and act on the error. Thanks @Mark-Life (#774)
--- a/packages/core/api/src/handlers/oauth.ts
+++ b/packages/core/api/src/handlers/oauth.ts
@@ -8,7 +8,7 @@ import { HttpApiBuilder } from "effect/unstable/httpapi";
 import { HttpServerResponse } from "effect/unstable/http";
 import { Effect, Option, Predicate, Schema } from "effect";
 
-import { runOAuthCallback } from "../oauth-popup";
+import { runOAuthCallback, type PopupErrorMessage } from "../oauth-popup";
 import {
   OAUTH_POPUP_MESSAGE_TYPE,
   OAuthCompleteError,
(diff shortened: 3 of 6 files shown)


RhysSullivan:
thanks!

PR #799, opened 2026-05-13
fix: align CLI and Desktop local runtimes

bheemreddy-samsara:
## Summary

CLI commands and Executor Desktop now avoid split-brain local runtimes in both startup orders.

If Desktop is already running, CLI commands can target its authenticated sidecar with `--base-url` or `EXECUTOR_BASE_URL`. Credentials embedded in that URL are sent as Basic-auth headers and stripped from display output. If a local endpoint returns `401 Unauthorized`, the CLI now explains that credentials are required instead of treating the occupied port as unavailable and starting a second daemon.

If a compatible CLI daemon is already running first, Desktop discovers it from the CLI daemon pointer files and attaches only when Desktop auth is disabled and the daemon serves the shared `~/.executor-global` scope. Authenticated, stale, malformed, or different-scope listeners are skipped so Desktop does not attach to an unrelated runtime or bypass its configured auth settings.

Desktop uses `~/.executor-global` as its default shared scope so MCPs configured for CLI/agent usage are visible in Desktop too. Desktop settings are also initialized lazily after Electron sets its `userData` path, preventing Desktop from reading stale settings from Electron's default app-data location and accidentally spawning on the CLI port.

## Testing

- `bunx vitest run tests/desktop-server-discovery.test.ts tests/daemon-bootstrap.test.ts`
- `bun run --cwd apps/desktop typecheck`
- `bun run --cwd apps/cli typecheck`
- `bun run format:check`
- `bun run lint`
- `bunx --bun electron-vite build` from `apps/desktop`
- `CSC_IDENTITY_AUTO_DISCOVERY=false bunx electron-builder --mac dir --arm64 --config electron-builder.config.ts` from `apps/desktop`
- Manual packaged Desktop app launch confirmed the app uses the shared `~/.executor-global` scope and syncs the MCP sources from `~/.executor-global/executor.jsonc`.
- Manual packaged Desktop app launch confirmed unauthenticated `/api/scope` returns `401` and authenticated `/api/scope` returns `200` on the Desktop sidecar when it owns port `4789`.
- Manual 401 probe with a local server confirmed the CLI reports the auth requirement instead of starting a duplicate daemon.
- Manual authenticated `daemon status --base-url http://executor:secret@...` probe confirmed Basic auth headers are sent and output hides credentials.
- Manual local discovery probe with a real HTTP server returning the shared scope confirmed Desktop discovery selects the existing runtime when given a verified candidate.

### DIFF
--- a/.changeset/cli-authenticated-base-url.md
+++ b/.changeset/cli-authenticated-base-url.md
@@ -1,0 +1,6 @@
+---
+"executor": patch
+"@executor-js/desktop": patch
+---
+
+Allow CLI commands to target authenticated local runtimes with `--base-url` or `EXECUTOR_BASE_URL`, report local 401 responses instead of starting a duplicate daemon, and let Desktop attach to an existing unauthenticated CLI daemon using the shared `~/.executor-global` scope.
--- a/README.md
+++ b/README.md
@@ -93,6 +93,16 @@ executor call gmail send '{"to":"alice@example.com","subject":"Hi"}'
 `executor call`, `executor resume`, and `executor tools ...` commands auto-start a local daemon if needed.
 If the default port is busy, the CLI will pick an available local port and track it automatically.
 
+To point CLI commands at an existing authenticated local runtime, pass `--base-url` or set `EXECUTOR_BASE_URL`:
+
+```bash
+EXECUTOR_BASE_URL='http://executor:<password>@127.0.0.1:4789' executor tools sources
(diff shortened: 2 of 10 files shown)


bheemreddy-samsara:
@RhysSullivan thoughts here ?

RhysSullivan:
ended up fixing this one w/ a different attachment method

PR #817, opened 2026-05-14
[codex] Cut storage over to FumaDB

RhysSullivan:
## Summary

- Vendor FumaDB into the workspace and make Executor storage speak FumaDB tables directly.
- Remove the core SDK Drizzle/PGlite entrypoints so SDK consumers pass a FumaDB database instead of inheriting a bundled database choice.
- Keep the local app on SQLite via a FumaDB SQLite composition, with a one-shot import path for existing local SQLite data.
- Move Postgres/PGlite setup into the cloud and worker test/dev owners that actually need it.
- Move runtime Drizzle schema setup into the vendored FumaDB adapter layer.

## Impact

This is a full storage cutover. The old custom storage model and SDK-owned database helpers are removed rather than kept as compatibility layers. Core stays database-agnostic at runtime, while apps/tests compose the FumaDB database they want.

## Coverage Notes

- Restores local legacy Drizzle migration coverage for OpenAPI, MCP, GraphQL, Google Discovery, OAuth connection normalization, and migration SQL nesting.
- Adds local DB upgrade and SQLite-to-Fuma import coverage, including plugin-shaped legacy rows with JSON, boolean, bigint, date, nullable, scope, and id values.
- Adds Fuma-native executor coverage for transaction rollback, unique violation propagation, close hooks, and source detection behavior.
- Restores MCP RFC 9728 protected-resource metadata probe coverage.

## Validation

- `bun run format:check`
- `bun run lint`
- `bun run scripts/check-changelog-stubs.ts`
- `bun run --cwd apps/local typecheck`
- `bun run --cwd packages/core/sdk typecheck`
- `bun run --cwd packages/plugins/mcp typecheck`
- `bun run --cwd apps/local test -- src/server/db-upgrade.test.ts src/server/sqlite-import.test.ts src/server/mcp-oauth.test.ts src/server/migrate-openapi-bindings.test.ts src/server/migrate-mcp-bindings.test.ts src/server/migrate-graphql-bindings.test.ts src/server/migrate-google-discovery-bindings.test.ts src/server/migrate-oauth-connections.test.ts src/server/migration-nesting.test.ts`
- `bun run --cwd packages/core/sdk test -- src/executor.test.ts`
- `bun run --cwd packages/plugins/mcp test -- src/sdk/probe-shape.test.ts`
- `node ../../node_modules/vitest/vitest.mjs run src/mcp-flow.test.ts` from `apps/cloud`


packages/plugins/openapi/src/sdk/store.ts
```diff
@@ -396,25 +370,79 @@ export interface OpenapiStore {
...
+type StoreConditionBuilder = ConditionBuilder<Record<string, AnyColumn>>;
+
+const bySourceScope =
+  (sourceId: string, scope: string) =>
+  (b: StoreConditionBuilder): Condition =>
+    b.and(b("source_id", "=", sourceId), b("scope_id", "=", scope)) as Condition;
+
+const byScopedId =
+  (id: string, scope: string) =>
+  (b: StoreConditionBuilder): Condition =>
```
RhysSullivan: this is all implemented really weirdly, i.e why not just use fumadb like normal and why is every plugin redeclaring all of these and there's no type safety

apps/cli/src/build.ts
RhysSullivan, line 1: why'd we remove embedded migrations

RhysSullivan: This was removed because the local app no longer runs Drizzle migrations from disk. Local now builds the FumaDB SQLite schema from `collectTables(...)` at startup via FumaDB/Drizzle runtime schema generation, and the legacy SQLite cutover imports rows directly from the old DB into the new Fuma DB. So there is no migration folder for the CLI binary to embed anymore; keeping the old embedded Drizzle migration module would be dead code.

PR #821, opened 2026-05-14
Add FumaDB table policy interceptors

RhysSullivan:
Stacked on #817.

## Summary
- add typed FumaDB table policy hooks for read/create/update/delete requests
- enforce executor scope policy by constraining FumaDB query conditions instead of prefetching rows
- add realistic scope-policy coverage for buggy plugin reads, counts, updates, and deletes

## Verification
- bun run format:check
- bun run lint
- bun run typecheck
- bun run --cwd packages/core/sdk test -- src/scope-policy.test.ts src/credential-bindings.test.ts
- bun run --cwd apps/local test -- src/server/sqlite-import.test.ts
- bun run --cwd packages/plugins/openapi test -- src/sdk/plugin.test.ts

### DIFF
--- a/apps/local/src/server/sqlite-import.test.ts
+++ b/apps/local/src/server/sqlite-import.test.ts
@@ -4,7 +4,8 @@ import { existsSync, mkdtempSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
-import { collectTables, withQueryContext } from "@executor-js/sdk";
+import { collectTables } from "@executor-js/sdk";
+import { withQueryContext } from "fumadb/query";
 
 import { importSqliteDataToFuma } from "./sqlite-import";
 import { createSqliteFumaDb, type SqliteFumaDb } from "./sqlite-fumadb";
--- a/apps/local/src/server/sqlite-import.test.ts
+++ b/apps/local/src/server/sqlite-import.test.ts
@@ -81,10 +81,11 @@ describe("importSqliteDataToFuma", () => {
       path: join(workDir, "target.db"),
     });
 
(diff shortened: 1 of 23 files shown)


apps/local/src/server/sqlite-import.test.ts
```diff
@@ -96,7 +96,9 @@ describe("importSqliteDataToFuma", () => {
     expect(existsSync(sqlitePath)).toBe(false);
     expect(result.backupPath && existsSync(result.backupPath)).toBe(true);
 
-    const source = (await sqlite.db.findFirst("source", {
+    const db = withQueryContext(sqlite.db, { allowedScopeIds: new Set(["scope_a"]) });
```
RhysSullivan: are we giving raw db access that bypasses the filtering we have

apps/local/src/server/sqlite-import.ts
```diff
@@ -180,8 +186,11 @@ export const importSqliteDataToFuma = async (
     sqlite = new Database(options.sqlitePath, { readonly: true });
     const importedTables: string[] = [];
     let importedRows = 0;
+    const dbWithScopeContext = withQueryContext(options.db, {
```
RhysSullivan: options.db probs shouldnt exist

packages/core/api/src/scoped-targets.test.ts
```diff
@@ -175,7 +176,9 @@ describe("core API explicit target scopes", () => {
       expect(response.status).toBe(200);
       const rows = toScopeRows(
         yield* Effect.promise(() =>
-          config.db.findMany("connection", {
+          withQueryContext(config.db, {
```
RhysSullivan: why isnt .db already scoped

PR #824, opened 2026-05-15
fix(oauth): surface real callback error in popup with collapsible details

### LINKED ISSUE #773: [UX] OAuth callback errors are swallowed; popup always shows generic "Authentication failed"

Mark-Life:
## Summary
- Drop the `Effect.catchCause` wrapper in the OAuth callback handler so the original tagged error propagates with its real message instead of a hardcoded "Authentication failed" placeholder.
- Popup now renders a short, tag-derived headline plus the full technical message inside a collapsible `<details>` disclosure.
- Skip auto-close on failure so users can read and act on the error.
- Backwards-compatible: existing consumers reading `result.error` keep working.

Closes #773

## Test plan
- [ ] `bun run test` for `packages/core/api` (new `oauth-popup.test.ts` cases)
- [ ] Manual: trigger an OAuth callback failure; verify headline + collapsible details render and popup stays open
- [ ] Manual: trigger an OAuth success; verify existing flow/auto-close unchanged

### DIFF
--- a/.changeset/oauth-popup-error-details.md
+++ b/.changeset/oauth-popup-error-details.md
@@ -1,0 +1,5 @@
+---
+"executor": patch
+---
+
+Surface the real OAuth callback error in the popup with a collapsible details disclosure. See `apps/cli/release-notes/next.md`.
--- a/apps/cli/release-notes/next.md
+++ b/apps/cli/release-notes/next.md
@@ -11,3 +11,7 @@ Two regressions kept `executor.jsonc` and the runtime DB from agreeing on which 
 ### Variadic tool path arguments no longer crash
 
 Calling a tool with multiple positional path arguments (`executor <tool> path/a path/b ...`) no longer panics in the CLI argument parser. Thanks @grfwings (#761)
+
+### OAuth popup surfaces the real callback error
+
+OAuth callback failures previously rendered a hardcoded `"Authentication failed"`, hiding the actual cause behind a generic placeholder. The popup now shows a short tag-derived headline plus the full technical message inside a collapsible `<details>` disclosure, and skips auto-close on failure so users can read and act on the error. Thanks @Mark-Life (#774)
--- a/packages/core/api/src/handlers/oauth.ts
+++ b/packages/core/api/src/handlers/oauth.ts
@@ -8,7 +8,7 @@ import { HttpApiBuilder } from "effect/unstable/httpapi";
 import { HttpServerResponse } from "effect/unstable/http";
 import { Effect, Option, Predicate, Schema } from "effect";
 
-import { runOAuthCallback } from "../oauth-popup";
+import { runOAuthCallback, type PopupErrorMessage } from "../oauth-popup";
 import {
   OAUTH_POPUP_MESSAGE_TYPE,
   OAuthCompleteError,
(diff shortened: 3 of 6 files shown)


Mark-Life:
Closing — opened by mistake, change is already merged upstream.
sorry for this. i forgot how to upstream property and did a wrong command

PR #825, opened 2026-05-15
[codex] Use compiler-backed schema previews

RhysSullivan:
## Summary

- Replace the hand-written JSON Schema to TypeScript preview renderer with compiler-backed generation.
- Vendor the compiler into `@executor-js/json-schema-to-typescript` so Executor owns the schema compiler used by `@executor-js/sdk`.
- Remove the compiler's Prettier dependency; the vendored formatter is now a no-op and callers normalize preview display themselves.
- Fix vendored compiler type-only imports/exports so Bun runtime imports do not fail and collapse previews to `unknown`.
- Normalize refs and nullable schemas before compilation so code-mode previews get valid TypeScript for dashed refs, quoted dashed properties, nullable enums/constants, recursive refs, and large schema graphs.
- Return compact schema roots plus reachable shared definitions, instead of embedding full `$defs` into every root.
- Keep the browser bundle off Node-only SDK/compiler code through `@executor-js/sdk/shared`.
- Make the local UI schema tab load JSON schema only; TypeScript compilation now runs only when `includeTypeScript=true` / the TypeScript tab is opened.
- Trim OpenAPI source metadata and tools-list responses so Cloudflare source pages do not download raw specs or all row descriptions up front.

## Root Cause

The previous preview renderer built TypeScript strings by hand and used raw `$ref` tails as type names, so schema names such as `foo-bar` could produce invalid TypeScript. Adding a real compiler at the SDK root also exposed an existing package-boundary issue: React imported `@executor-js/sdk` for shared IDs/errors/policy helpers, which could pull Node-only compiler dependencies into the browser bundle.

The remaining Cloudflare load slowness had three separate causes: the source detail page fetched the full stored OpenAPI spec even though it only needed metadata, the default schema tab eagerly ran the TypeScript compiler despite rendering the raw schema view first, and the tool-list endpoints shipped descriptions/annotation text for every row before any tool was selected.

After vendoring, Bun's TypeScript runtime hit upstream-style type-only imports/exports emitted as runtime imports. The SDK caught that import failure and returned `unknown` for the whole preview. The vendored package now uses explicit `import type` / `export type` for those symbols.

## Vendored Compiler Notes

- `@executor-js/sdk` now depends on `@executor-js/json-schema-to-typescript` via `workspace:*`.
- The vendored compiler is based on `json-schema-to-typescript@15.0.4` source, without the upstream CLI entrypoint.
- Runtime deps for the vendored compiler are only `@apidevtools/json-schema-ref-parser`, `js-yaml`, and `lodash`.
- Prettier can still appear in the root lockfile through unrelated repo tooling and transitive dependencies, but it is no longer part of the Executor SDK schema compiler path.

## Local Measurements

- Cloudflare OpenAPI source metadata: ~11,031,966 B before, 317 B after omitting `config.spec`.
- Cloudflare source tools list: ~1,160,557 B before, 258,518 B after compact rows.
- Cloudflare global tools list: ~1,096,457 B before, 237,667 B after compact rows.
- Cloudflare schema endpoint, default JSON/schema view warm runs: 71,222 B, ~0.06-0.14s.
- Cloudflare schema endpoint with `includeTypeScript=true` after runtime import fix: Input `{ account_id: AccessIdentifier; body: AccessAppRequest; }`, Output `(AccessAppsComponentsSchemasSingleResponse & { result?: AccessAppResponse; })`, 184 shared TypeScript definitions.

## Validation

- `bun run --cwd packages/core/json-schema-to-typescript typecheck`
- `bun run --cwd packages/core/json-schema-to-typescript build`
- `bun run --cwd packages/core/sdk test src/schema-types.test.ts`
- Bun runtime repro for `buildToolTypeScriptPreview` returns `{ account_id: string; body: unknown; }` instead of root `unknown`.
- Live Cloudflare schema endpoint repro on a fresh local server returns `{ account_id: AccessIdentifier; body: AccessAppRequest; }`.
- `bun run --cwd packages/core/sdk test -- src/schema-refs.test.ts`
- `bun run --cwd packages/core/sdk test -- src/executor.test.ts`
- `bun run --cwd packages/core/execution test -- src/tool-invoker.test.ts`
- `bun run --cwd packages/core/sdk typecheck`
- `bun run --cwd packages/core/api typecheck`
- `bun run --cwd packages/react typecheck`
- `bun run --cwd packages/plugins/openapi typecheck`
- `bun run --cwd packages/plugins/mcp typecheck`
- `bun run --cwd packages/plugins/graphql typecheck`
- `bun run --cwd packages/plugins/google-discovery typecheck`
- `bun run --cwd packages/plugins/onepassword typecheck`
- `bun run --cwd packages/core/sdk build`
- `bun run --cwd apps/local build`
- `bunx oxfmt --check` on changed files
- `git diff --check`
- `oxlint --deny-warnings` on changed TypeScript files


apps/cli/src/main.ts
```diff
@@ -1106,7 +1107,7 @@ const runCallHelp = (
       exactTool: exactTool
         ? {
             id: exactTool.id,
-            description: exactTool.description,
+            description: exactToolSchema?.description,
```
RhysSullivan: why's this nullish now

apps/cli/src/main.ts
```diff
@@ -1065,28 +1065,29 @@ const runCallHelp = (
       ? tools.find((tool) => tool.id === inspection.exactPath)
       : undefined;
 
-    if (exactTool && inspection.children.length === 0) {
-      const schema = yield* client.tools
-        .schema({
```
RhysSullivan, line 1070: why'd this conditional get removed./ moved

apps/cloud/src/services/sources-refresh.node.test.ts
```diff
@@ -123,8 +123,8 @@ describe("sources.refresh (HTTP)", () => {
           }),
         );
         expect(beforeTools.length).toBe(1);
-        expect(beforeTools.some((t) => t.name.startsWith("ping"))).toBe(true);
-        expect(beforeTools.some((t) => t.name.startsWith("pong"))).toBe(false);
+        expect(beforeTools.some((t) => t.id.startsWith(`${namespace}.ping.`))).toBe(true);
+        expect(beforeTools.some((t) => t.id.startsWith(`${namespace}.pong.`))).toBe(false);
```
RhysSullivan: why are we going from id -> name

apps/cloud/src/services/sources-refresh.node.test.ts
```diff
@@ -123,8 +123,8 @@ describe("sources.refresh (HTTP)", () => {
           }),
         );
         expect(beforeTools.length).toBe(1);
-        expect(beforeTools.some((t) => t.name.startsWith("ping"))).toBe(true);
-        expect(beforeTools.some((t) => t.name.startsWith("pong"))).toBe(false);
+        expect(beforeTools.some((t) => t.id.startsWith(`${namespace}.ping.`))).toBe(true);
```
RhysSullivan: why is this prefixed with namespace now

apps/cloud/src/services/tenant-isolation.node.test.ts
```diff
@@ -140,7 +140,6 @@ describe("tenant isolation (HTTP)", () => {
       const orgBTools = yield* asOrg(orgB, (client) =>
         client.tools.list({ params: { scopeId: ScopeId.make(orgB) } }),
       );
-      expect(orgBTools.map((t) => t.sourceId)).not.toContain(namespaceA);
       for (const id of orgBTools.map((t) => t.id)) {
```
RhysSullivan, line 144: why'd this get removed

PR #837, opened 2026-05-16
fix: parse quoted stdio env values

### LINKED ISSUE #811: stdio MCP env values keep literal quotes (parseStdioEnv lacks dotenv parsing)
## Repro
Add an stdio MCP source via the UI with dotenv-style env, e.g.:

```
IBKR_HOST="127.0.0.1"
IBKR_PORT="4002"
```

## Behavior
Child process receives `IBKR_HOST` = `"127.0.0.1"` (literal 11-char string with quotes), schema/Config validation in the child fails, child exits, and the API returns `McpToolDiscoveryError` surfaced as a generic "Failed connecting via stdio".

## Cause
`parseStdioEnv` in `packages/plugins/mcp/src/react/AddMcpSource.tsx:516` does a naive `=`-split + `.trim()` — it does not interpret surrounding quotes as syntactic.

Secondary: `connectClient` in `packages/plugins/mcp/src/sdk/connection.ts:88-94` discards the underlying error, so the UI can't show the real cause.

## Fix
Parse env like dotenv: matched surrounding `"…"` or `'…'` are treated as delimiters (with the usual escape handling inside double quotes). Optionally also propagate the underlying error message into `McpConnectionError` for better debuggability.

asim48-ctrl:
## Summary
- strip matching single or double quotes from stdio MCP env values
- support common double-quoted escapes so pasted dotenv-style snippets pass clean values to child processes

Fixes #811.

## Validation
- `git diff --check`
- local parser smoke check for quoted host/port values, single-quoted literals, and double-quoted escapes

### DIFF
--- a/packages/plugins/mcp/src/react/AddMcpSource.tsx
+++ b/packages/plugins/mcp/src/react/AddMcpSource.tsx
@@ -513,13 +513,43 @@ export default function AddMcpSource(props: {
     return args;
   };
 
+  const parseStdioEnvValue = (raw: string): string => {
+    const value = raw.trim();
+    if (value.length < 2) return value;
+
+    const quote = value[0];
+    if ((quote !== '"' && quote !== "'") || value[value.length - 1] !== quote) {
+      return value;
+    }
+
+    const inner = value.slice(1, -1);
+    if (quote === "'") return inner;
+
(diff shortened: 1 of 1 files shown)


RhysSullivan:
thanks

PR #876, opened 2026-05-30
fix: declare effect as a peer dependency in published packages

aryasaatvik:
## Problem

The published `@executor-js/*` packages declare `effect` as a hard `dependency`. Since `effect` needs a **single module instance** for `Context`/service identity and type equality, a consumer pinning a different `4.0.0-beta.x` ends up with a duplicated `effect` — services stop resolving, and TypeScript flags the two copies as incompatible.

## Fix

Move `effect` from `dependencies` → `peerDependencies` (with a `devDependencies` mirror) in the 12 published library packages, so the consumer supplies the single shared instance. Matches how `react` / `@effect/atom-react` are already declared as peers.

The `executor` CLI (leaf binary) and private workspace packages keep `effect` as a regular dependency.

## Verification

- `bun install` → single `effect@4.0.0-beta.59`
- `turbo run typecheck` → 34/34 pass
- `bun pm pack` on `@executor-js/sdk` → `effect` under `peerDependencies`, not `dependencies`

### DIFF
--- a/.changeset/effect-peer-dependency.md
+++ b/.changeset/effect-peer-dependency.md
@@ -1,0 +1,9 @@
+---
+"executor": patch
+---
+
+Move `effect` from `dependencies` to `peerDependencies` (with a `devDependencies` mirror) in the published library packages.
+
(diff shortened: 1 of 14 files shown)


RhysSullivan:
Superseded by #922, which reapplied this on current main. Thanks @aryasaatvik.

PR #878, opened 2026-06-01
Add org-scoped MCP routes for the cloud app

RhysSullivan:
Let an MCP client pin a specific organization in the URL (`/<org_id>/mcp`)
instead of relying on the organization claim in the OAuth access token. This
unblocks clients that hold a single sign-in session but need to target one of
several organizations.

Server:
- Serve `/<org_id>/mcp` and the matching
  `/.well-known/oauth-protected-resource/<org_id>/mcp` discovery document
  alongside the existing bare `/mcp`.
- Take the organization from the URL when present, falling back to the token
  claim otherwise. Membership is verified against the auth provider on every
  request, so the URL is a selector, not a trust boundary.
- Consolidate MCP path ownership behind one predicate shared by the request
  middleware, the handler, and the test worker. Only `org_`-shaped segments are
  claimed, so unrelated `/<segment>/mcp` paths still fall through to routing.

Frontend:
- The cloud "Connect your MCP client" page and the install card now hand out the
  org-scoped URL. The shared install card stays unscoped for local/desktop,
  where there is no organization.

Tests cover org-scoped discovery metadata, URL-supplied organization, membership
denial before session creation, non-org paths falling through, and org-scoped
URL/command rendering.

Draft: interim org-id form; migrates to organization handles once URL-based org
routing lands.

### DIFF
--- a/apps/cloud/src/mcp-flow.test.ts
+++ b/apps/cloud/src/mcp-flow.test.ts
@@ -98,19 +98,26 @@ type McpPostInit = {
   readonly accept?: string;
 };
 
-const mcpPost = (init: McpPostInit): Promise<Response> => {
+const mcpPostTo = (url: string, init: McpPostInit): Promise<Response> => {
   const headers: Record<string, string> = {
     "content-type": CONTENT_TYPE_JSON,
     accept: init.accept ?? JSON_AND_SSE,
   };
   if (init.bearer) headers.authorization = `Bearer ${init.bearer}`;
   if (init.sessionId) headers["mcp-session-id"] = init.sessionId;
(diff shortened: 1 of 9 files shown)


RhysSullivan:
* **#878** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/878?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/878?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #891, opened 2026-06-04
Emit named GraphQL operations for generated tools

oscnord:
## Problem

Generated GraphQL operations are anonymous — e.g. `query($slug: ID!) { contentcollection(slug: $slug) { ... } }`. Two consequences:

- Servers that require named operations reject every invocation. (Apollo Server can be configured this way; some gateways enforce it unconditionally and return a 400 before the resolver runs.)
- APM/observability tooling (Apollo Studio, New Relic, Datadog) keys traces and transaction names off the operation name, so all anonymous operations collapse into a single unnamed bucket.

## Change

Derive an operation name from the root field and emit it in the generated document, so `contentcollection` produces:

```graphql
query Contentcollection($slug: ID!) { contentcollection(slug: $slug) { ... } }
```

- `buildOperationStringForField` and the anonymous fallback path both name the operation.
- The name is the root field name with its first character upper-cased; GraphQL field names are already valid name tokens, so no further escaping is needed.
- Each generated document contains exactly one operation, so there is no naming-collision concern.
- The request body is unchanged otherwise; servers derive `operationName` from the document, so no separate `operationName` field is required.

## Tests

Added a test asserting a query and a mutation are sent as named operations (`query Hello …`, `mutation SetGreeting …`). Full `@executor-js/plugin-graphql` suite passes; typecheck, oxfmt, and oxlint are clean.

### DIFF
--- a/.changeset/named-graphql-operations.md
+++ b/.changeset/named-graphql-operations.md
@@ -1,0 +1,5 @@
+---
+"executor": patch
+---
+
+GraphQL sources now emit named operations (e.g. `query Hello { ... }`) instead of anonymous ones. This fixes invocation against servers that reject anonymous operations, and gives APM tooling that keys on the operation name a meaningful value. The operation name is derived from the root field name.
--- a/.changeset/named-graphql-operations.md
+++ b/.changeset/named-graphql-operations.md
@@ -1,5 +1,5 @@
 ---
-"executor": patch
+"@executor-js/plugin-graphql": patch
 ---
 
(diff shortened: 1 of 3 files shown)


RhysSullivan:
Rebased this onto current main and fixed the stale test setup plus changeset target.\n\nValidation run locally on the rebased branch:\n- `bun run --cwd packages/plugins/graphql test`\n- `bun run --cwd packages/plugins/graphql typecheck`\n- `bunx changeset status --since=origin/main`\n- `bun run format:check -- .changeset/named-graphql-operations.md packages/plugins/graphql/src/sdk/plugin.ts packages/plugins/graphql/src/sdk/plugin.test.ts`

PR #935, opened 2026-06-10
Fix desktop sidecar packaging and release pipeline

RhysSullivan:
The desktop sidecar build was missing a few pieces that the CLI build already had, which meant recent desktop builds couldn't start their local server. This brings the two builds back in line and tightens up the release workflow so this class of issue gets caught before publishing.

## Changes

**Sidecar packaging**
- Stage the `libsql.node` and `keyring.node` native bindings next to the compiled sidecar binary (`bun build --compile` doesn't bundle `.node` addons). A new `native-bindings.ts` first-import points the loaders at them, mirroring the CLI build.
- Embed the legacy v1 migration chain (`drizzle-legacy-v1`) in the sidecar instead of the v2 chain, so the v1→v2 data migration can replay it against older databases.

**Release workflow**
- Run the compiled-sidecar smoke test (`test:smoke`) in the publish workflow on legs whose target arch matches the runner. This exercises the compiled binary end to end (native bindings, embedded migrations, MCP → QuickJS → OpenAPI round trip), which dev mode can't cover.
- The two mac matrix legs each emit a `latest-mac.yml` that only lists their own arch, and the release job previously kept whichever uploaded last. They're now renamed per arch and merged into a single manifest (`scripts/merge-latest-mac-yml.ts`) so the updater serves the right artifact to each arch.
- Drop the config-level arch pins for win/linux electron-builder targets — the workflow's `--x64`/`--arm64` flag now decides what gets built, so an installer only ships a sidecar compiled for its own arch.

**Startup UX**
- When the sidecar fails to start, show an error dialog with the underlying error and log path instead of quitting silently, and check for updates with install-on-quit enabled so an affected install can pick up a fixed build on its own.

**Smoke test refresh**
- The fixture now replays the real legacy migration chain (instead of a hand-rolled partial schema), asserts the v1→v2 backup, and drives the current addSpec → connection → tool-invocation flow. Failures throw instead of exiting so the spawned sidecar is always torn down.

## Testing
- `bun run test:smoke` (desktop) green against the compiled binary, including the v1→v2 migration path.
- `scripts/merge-latest-mac-yml.ts` verified against the manifests from an existing release.
- Packaged the mac arm64 app locally and confirmed it launches and serves the UI.
- `typecheck`, `lint`, and `format:check` green on the touched packages.

### DIFF
--- a/.github/workflows/publish-desktop.yml
+++ b/.github/workflows/publish-desktop.yml
@@ -29,22 +29,29 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # smoke: run the compiled-sidecar smoke test on legs whose target
+          # matches the runner (the mac x64 leg cross-compiles on an arm64
+          # runner, so its binary can't be executed natively there).
           - os: macos-latest
             arch: arm64
             platform: mac
             bun-target: bun-darwin-arm64
+            smoke: true
           - os: macos-latest
             arch: x64
             platform: mac
             bun-target: bun-darwin-x64
+            smoke: false
           - os: ubuntu-latest
             arch: x64
             platform: linux
             bun-target: bun-linux-x64
+            smoke: true
           - os: windows-latest
             arch: x64
             platform: win
             bun-target: bun-windows-x64
+            smoke: true
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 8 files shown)


RhysSullivan:
* **#935** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/935?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/935?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #939, opened 2026-06-10
Fix desktop publish: cross-target install glob and Windows migration rename

RhysSullivan:
Two fixes for the desktop legs that failed in the v1.5.2 publish run (the new smoke gate did its job — neither would have been caught before publishing previously):

- **mac x64**: the cross-target native-deps install ran through Bun's `$` shell, which glob-expands the bare `*` in `--cpu=*` and fails with "no matches found". Switched to `Bun.spawn` with an argv array, matching `apps/cli/src/build.ts`.
- **Windows**: the compiled-sidecar smoke test caught the v1→v2 data migration failing with `EBUSY` when renaming `data.db` immediately after closing the SQLite handle — Windows releases file handles asynchronously (and runner antivirus takes short-lived locks), unlike POSIX where renaming an open file is fine. The file-set move now retries briefly on `EBUSY`/`EPERM`. This would have affected real Windows installs migrating v1 data, not just CI.

Includes a changeset (patch) so the fixed desktop artifacts ship as the next release; v1.5.2's desktop legs built from the tag and can't pick these up.

Testing: migration suite green; `BUN_TARGET=bun-darwin-x64` sidecar build now completes on an arm64 host and stages x86_64 bindings; arm64 `test:smoke` green; format/lint/typecheck green.

### DIFF
--- a/.changeset/desktop-publish-fixes.md
+++ b/.changeset/desktop-publish-fixes.md
@@ -1,0 +1,8 @@
+---
+"executor": patch
+---
+
+Desktop packaging follow-ups from the v1.5.2 release run:
+
+- Fixed the Intel mac desktop build failing in CI (the cross-target dependency install was being glob-expanded by the shell).
+- Fixed the first-launch data migration on Windows: renaming the previous database file could hit a transient `EBUSY` while the just-closed SQLite handle was released, so the move now retries briefly instead of failing startup.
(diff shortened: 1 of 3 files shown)


RhysSullivan:
* **#939** <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/939?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RhysSullivan/executor/939?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #941, opened 2026-06-10
MCP: support query-parameter credential auth (e.g. ui.sh ?token=)

RhysSullivan:
## Summary

Adds a **query-parameter credential** auth method to the MCP plugin, so MCP servers that authenticate via a query-string token (`?token=<value>`) can be added and connected — e.g. ui.sh.

Follow-up to the multi-method auth work. The shared "+ Add method" modal already offers a **Query param** credential placement (the placement editor is plugin-agnostic), but MCP's auth model was only `none | header | oauth2` — no query kind. So picking "API key query" → `token` failed with **"Failed to add method"**: the codec dropped the query placement and never reached the API.

### Changes
- `McpAuthMethod` / `McpAuthMethodInput` gain `{ kind: "query", paramName, prefix? }`.
- `buildConnectorInput` renders the connection's resolved value into the endpoint's query string (`?<paramName>=<value>`), alongside the existing header/oauth paths.
- `describeMcpAuthMethods` projects it as an `apikey` method carrying a **query** placement (so the hub shows "API key (token)").
- React codec (editor↔method, placements→method, config→`AuthMethod`) and the edit-source label/diff handle the query carrier.

### Tests
- **Wire-level proof:** declare a query method → create a connection with a token → invoke a tool against a real test MCP server → assert the server received `?token=<value>` in the request URL and **no** Authorization header.
- describe / codec unit coverage for the query round-trip (the reproduction that returned `[]` now asserts the desired behavior).
- e2e scenario for the ui.sh shape: declare a `?token=` method and append one via the custom-method path — proving "Failed to add method" is gone.

## Test plan
- `bun run typecheck` (40/40), `bun run lint`, `bun run format:check` — green.
- `plugin-mcp` unit suite green; e2e `scenarios/auth-methods.test.ts` green on selfhost.


RhysSullivan:
Superseded: query-parameter credential auth is now supported through the shared placements auth model (landed via #933/#943). `queryParams` placements on apiKey methods cover the `?token=` case, with e2e coverage in the auth-methods scenarios. This branch only carried force-push-stale history on top of that, so closing and deleting it.

PR #952, opened 2026-06-10
Support per-call server selection for OpenAPI tools

SferaDev:
Resolve OpenAPI servers per tool call instead of baking a single base URL at add time. Operations can declare multiple servers (document, path, or operation level) and templated URLs with `{variables}`; each is now selectable when invoking the tool.

## Changes

- **Per-call `server` input.** Each generated tool exposes an optional `server` input (`{ url?, variables? }`), shown only when there is something to choose: `url` becomes an enum when more than one server applies, and `variables` are drawn from the applicable servers. This mirrors the existing synthetic `contentType` input.
- **Invoke-time host resolution.** A connection base URL override wins when set; otherwise the call's chosen server (`server.url`, or the first applicable) is resolved with its `{variables}` (call values, else spec defaults); otherwise no host is prepended.
- **Base URL is now an optional override.** Off by default, required only when the spec declares no servers. The source form shows a combobox when there is more than one top-level server and a plain input otherwise, with the first server's resolved URL as the placeholder.
- **Removed connection-level `serverVariables`** in favor of per-call variables.

## Compatibility

Bindings persisted before this change keep working: a missing `servers` field is treated as empty and falls back to the connection base URL.

## End-to-end coverage

Two scenarios pin this behavior end to end:

- **API** — `e2e/scenarios/openapi-server-selection.test.ts`: a multi-server spec, added through the typed API with no base URL, produces a tool whose input schema exposes the per-call `server` selector — a `url` enum over the declared servers plus the templated server's `{variables}`. Runs on cloud and selfhost.
- **UI** — `e2e/scenarios/openapi-server-selection-ui.test.ts`: pasting a multi-server spec turns the add form's Base URL field into a picker over those servers and relabels it an optional override, with a hint that the host is otherwise chosen per tool call.

OpenAPI · the add form offers a server picker and an optional base URL for a multi-server spec (cloud)

### DIFF
--- a/e2e/scenarios/openapi-server-selection-ui.test.ts
+++ b/e2e/scenarios/openapi-server-selection-ui.test.ts
@@ -1,0 +1,72 @@
+// Cross-target (browser): the UI side of per-call server selection. When a
+// pasted spec declares more than one server, the Add OpenAPI source form turns
+// the Base URL field into a picker over those servers and relabels it an
+// optional override — the host is otherwise resolved per tool call. The session
+// video + per-step screenshots are the artifact; this scenario skips on targets
+// without a browser surface (selfhost today).
+import { expect } from "@effect/vitest";
+import { Effect } from "effect";
+
+import { scenario } from "../src/scenario";
+import { Browser, Target } from "../src/services";
+
+// Two servers — production and staging — so the form offers a base-URL picker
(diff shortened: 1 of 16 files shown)


RhysSullivan:
thanks!

PR #974, opened 2026-06-12
Org-slug console URLs across cloud, self-host, and cloudflare hosts

RhysSullivan:
## What

Console URLs now carry the organization slug on org-scoped hosts: `executor.sh/<org-slug>/policies` instead of `/policies`. One shared route tree serves every host — cloud, self-host, and the Cloudflare host canonicalize onto the active organization's slug, while local/desktop keep bare URLs.

Org URLs · console paths carry the organization slug (cloud)

## How

**Routing.** Every shared console route nests under an optional `{-$orgSlug}` segment, so the same TanStack tree matches both bare and slug-prefixed paths. The TanStack generator rewrites `createFileRoute()` ids on disk to match the mounted path, so all consumers must mount the shared files identically — each app now has a `tsr.routes.ts` read by both its vite config and a `routes:gen` script (programmatic `@tanstack/router-generator`), keeping CLI regens and dev builds in lockstep.

**Canonicalization.** A shared `OrgSlugGate` mounted inside each authenticated shell pins the URL to the active org's slug: bare URLs redirect in place (path + query preserved), unknown slugs snap back, and on cloud a foreign slug the caller is a member of switches the session into that org. The gate distinguishes "URL is foreign" from "session moved out from under the URL" (org create/switch elsewhere) — without that, the gate races the in-flight switch and rotates the single-use session refresh twice, invalidating the cookie. The URL slug is a selector, not a trust boundary; sessions scope every API call server-side.

**Slugs.** A shared vocabulary in `@executor-js/api` owns the grammar (2–48 chars of `[a-z0-9-]`), a reserved-segment list (`api`, `mcp`, `login`, console sections, marketing paths, `cdn-cgi`, …), name-derived slug generation with `-2`/`-3` suffixing on collision, and validation helpers. `AccountOrganization` (`/account/me`) now carries `slug`.

Per host:
- **Cloud**: `organizations.slug` unique column (+ migration and an idempotent backfill script — the runtime also mints slugs lazily on first resolution, so the backfill is pre-warming, not a deploy gate). Slugs are minted at org create / invite accept and are stable across renames so org URLs survive. The org switcher and create-org flows land on the new org's slug URL. Onboarding (`/create-org`, `/setup-mcp`) stays bare.
- **Self-host**: the seed previously looked the org up *by slug*, so changing `EXECUTOR_ORG_SLUG` would have silently forked the instance with a second org — it now adopts the single existing org and renames its slug. The env var is validated against the grammar at boot. `/join/<code>` invite links stay bare.
- **Cloudflare**: new `SELF_HOSTED_ORG_SLUG` var (validated), surfaced through the account provider.

## Testing

- New cross-target browser scenario `e2e/scenarios/org-slug-routing.test.ts` (recording above): `/account/me` advertises a grammar-valid slug, bare deep links canonicalize, unknown slugs snap back, in-shell navigation keeps the prefix. Green on cloud and selfhost.
- New unit suites: slug grammar/generation/reserved list; optional-segment router semantics (bare vs slugged matching, static-sibling precedence, href generation, param inheritance); console-route contract updated.
- Full e2e suites green: cloud 26/26 files (75 tests), selfhost 14/14 — including the org-switcher and org-limit scenarios updated for slug URLs.
- `format:check`, `lint`, `typecheck`, `test` all green.

## Deploy notes (cloud)

1. Apply the schema migration (adds the nullable unique `slug` column) before or with the deploy — old code ignores the column, new code lazily mints slugs, so ordering is forgiving.
2. Optionally run `db:backfill-org-slugs:prod` afterwards to pre-mint slugs in bulk.


RhysSullivan (commented):
dont merge without removing sequential ids from org slugify

RhysSullivan:
Rebased onto main (now includes the edge auth gate #975, fresh-worktree env #973, and the spec-blob stack). Two behavioral changes from the rebase:

- **Foreign slugs are now a 404, not a silent redirect.** Main's #975 added a real not-found page plus a test asserting unknown paths render it — under slug routing, single-segment typo paths (`/this-page-does-not-exist`) match the slugged index route, so silently canonicalizing them into the active workspace would have regressed that. A URL naming an org the session can't see now renders the not-found page (shared `OrgSlugNotFound`); cloud still switches the session when the caller is a member of the slug's org. The cross-target scenario asserts the 404 instead of the old snap-back.
- The auth-hint cookie introduced by #975 now carries the org `slug`, and the slug migration renumbered to `0003` (main took `0002`).

Gates re-run post-rebase: unit suites green, selfhost e2e 18/18, cloud e2e green (the `auth-methods` mixed-invocation and OpenCode-lifecycle scenarios flake under full-suite load — both pass focused and the auth-methods one reproduces on main's baseline).

RhysSullivan:
Rebased onto main again (PostHog instrumentation #987, render-shell-from-hint #986, hosted-client redirects #989). Two integration points from this round:

- The SSR auth hint now carries the org **slug** (minted lazily through the same user-store path) — main's edge gate renders the real shell from the hint on first paint, so the slug has to ride along or the canonicalizer has nothing to pin.
- The foreign-slug 404 / org-switch state renders **inside the shell chrome** via a new `content` slot on the shared shell, matching main's in-shell route 404 (and its test).

Also since the last update, per review feedback:
- **Slug collisions get a random 4-char discriminator** (`acme-corp-x7k2`), not `-2` ordinals.
- **MCP install URLs pin the org slug** (`/acme-corp/mcp`); the server accepts the slug and the legacy `/org_xxx/mcp` id form interchangeably (verified live: legacy URLs in existing agent configs keep working — same discovery echo, same membership enforcement).

Post-rebase verification: unit gates green; cloud e2e 32/34, selfhost 17/19 — the four failures are two scenarios × two targets, all `connect-handoff`, failing on a hosted `resend.emulators.dev` control-plane regression ("Values cannot be larger than…" on credential mint) that reproduces on main's baseline; unrelated to this branch.

Deploy order for cloud remains: apply migration `0003` (nullable unique `slug` column — old code ignores it, new code mints lazily, so ordering is forgiving), deploy, then optionally `db:backfill-org-slugs:prod` to pre-warm.

PR #975, opened 2026-06-12
Gate signed-out document requests at the edge instead of flashing the app-shell skeleton

RhysSullivan:
## Problem

The same signed-out visit to the cloud root, with the auth probe held open so the loading window is visible — `main` on the left, this branch on the right:

Signed-out visit, before vs after

Signed-out visitors to executor.sh/cloud were shown the authenticated app-shell skeleton (sidebar nav + content-card grid) until a client-side `/account/me` round trip returned 401 — at which point the page swapped to a login UI. Every unsigned-out load flashed the wrong shell.

## Fix

Three layers, each standing alone:

**1. SSR auth gate (`apps/cloud/src/auth/ssr-gate.ts`)**

A TanStack Start request middleware (last in the `requestMiddleware` chain in `start.ts`) that intercepts GET/HEAD document navigations before any React renders. It verifies the sealed `wos-session` cookie directly in the worker — unseal + JWT check against a cached JWKS — so there is no per-request WorkOS round trip except on token refresh. Signed-out visitors get a `302 /login?returnTo=<path>` before any app HTML is generated. An invalid cookie is cleared on the way out (on executor.sh, the cookie's mere presence routes `/` past the marketing page). WorkOS refresh tokens are single-use, so a rotated sealed session is re-set on the response.

`/cloud` (the path marketing CTAs link to) was previously a 404; it now redirects into the app.

**2. Auth-hint cookie (`packages/react/src/multiplayer/auth-hint.tsx`)**

A non-HttpOnly companion to the sealed session: a display-only Schema-validated snapshot of the authenticated identity. `AuthProvider` writes it whenever `/account/me` confirms a session and reads it back at mount to seed optimistic "authenticated" state — so signed-in loads paint the real shell immediately without waiting for the probe. Cloud logout and the SSR gate clear it server-side. The resolved `/account/me` always wins.

**3. `/login` route with redirect-back**

A proper `/login` page (`apps/cloud/src/routes/app/login.tsx`) with `?returnTo=` round-trip plumbing: the SSR gate sets `returnTo`, the login handler encodes it into the OAuth `state` parameter beside the CSRF nonce (`base64url(JSON { nonce, returnTo })` — no extra cookie), and the callback decodes it from the validated state and redirects there. A `isSafeReturnTo` validator (relative-only, no `//`, no `/api`) prevents open-redirect abuse at both ends. Signed-in visitors hitting `/login` are bounced straight back.

`notFoundComponent` added — unknown paths now show a real 404 page instead of a skeleton or a blank app.

## Test coverage

`e2e/cloud/unauthenticated-skeleton.test.ts` — 4 scenarios:

1. Signed-out root → `/login` immediately with no app-shell flash (screenshot verified during the in-flight window)
2. Deep-link returnTo round trip (`/tools` → gate → login → callback → `/tools`) + open-redirect rejection
3. Signed-in passthrough + `/login` bounce
4. Real 404 page for unknown paths

`e2e/cloud/auth-hint.test.ts` — 2 scenarios: a confirmed `/account/me` writes the hint and the next load paints the real shell while the probe is held open; logout clears the hint with the session.

`e2e/cloud/session-gate.test.ts` — 3 scenarios: an invalid `wos-session` is cleared on the way to `/login` (along with the hint); `/cloud` routes into the app signed out and signed in; and a session whose access token no longer verifies is refreshed in-flight against the WorkOS emulator — the rotated sealed session reaches the browser, the spent one is revoked server-side (real single-use refresh tokens), and the rotated one keeps working. The stale-token path unseals the real cookie with the same iron-webcrypto sealing the WorkOS SDK uses, corrupts the JWT signature, and reseals — exercising the refresh branch without waiting out a real expiry.

`apps/cloud/src/auth/return-to.test.ts` + `login-state.test.ts` — unit tests covering the `isSafeReturnTo` / `safeReturnTo` / `loginPath` validators and the OAuth state envelope codec (round-trip, URL-safe alphabet, junk-tolerance).

All gates pass: `bun run format:check`, `bun run lint`, `bun run typecheck`, `bun run test`.

### DIFF
--- a/packages/react/src/multiplayer/auth-context.tsx
+++ b/packages/react/src/multiplayer/auth-context.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, useContext, useEffect } from "react";
+import React, { createContext, useContext, useEffect, useMemo, useState } from "react";
 import { useAtomValue } from "@effect/atom-react";
 import * as AsyncResult from "effect/unstable/reactivity/AsyncResult";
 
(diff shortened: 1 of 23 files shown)


packages/react/src/multiplayer/auth-context.tsx
```diff
@@ -82,6 +120,21 @@ const AuthProviderClient = ({
...
+    if (status === "authenticated" && userId && email !== null) {
+      writeAuthHintCookie({
+        v: 1,
+        user: { id: userId, email, name, avatarUrl },
+        organization: organizationId ? { id: organizationId, name: organizationName ?? "" } : null,
+      });
+    } else if (status === "unauthenticated") {
+      clearAuthHintCookie();
+    }
+  }, [status, userId, email, name, avatarUrl, organizationId, organizationName]);
```
RhysSullivan: implement this like someone who actually knows how to write react

RhysSullivan: Reworked in 16385edb9 — the resolved state is now a single `useMemo` keyed on the atom value (which is referentially stable between query emissions), so both effects depend on `resolved` directly instead of seven hand-destructured primitives. The identify effect passes the resolved state through as-is, the hint effect writes `resolved.user`/`resolved.organization` without reassembling them, and the context value is memoized.

PR #1001, opened 2026-06-13
Document Pi bridge

gvkhosla:
Adds a small README section for using Executor from Pi.

Pi does not currently include a built-in MCP client, so this points users to the community Pi extension that bridges Pi to `executor mcp` and includes the basic verification steps.

Docs-only change.


RhysSullivan:
thanks!

PR #1019, opened 2026-06-14
cli: service install takes over a running daemon instead of refusing (seamless upgrade)

RhysSullivan:
## The bug (upgrade flow)

`service install` is the upgrade path — `service status` literally tells users to
run it on version drift — but it **refused** whenever any daemon already owned
the data dir:

```
$ executor service install --port 4789
A local Executor cli-daemon is already running at http://localhost:4789 (pid 631).
Stop it first (quit the desktop app, or `executor daemon stop`), then re-run install.
   → exit 1; predecessor keeps the port; no service installed
```

On a real upgrade (`npm i -g executor@latest && executor service install`, or the
desktop app auto-installing the service) the user has **no idea which pid to
kill**, so the upgrade just fails. Reproduced in a real macOS VM (before this
change): `install exit=1`, predecessor still owns `:4789`.

## The fix — take over, don't refuse

Whatever holds the data dir (a foreground `executor web`, a manual `daemon run`,
or an old desktop sidecar) writes the **same** `server-control` manifest, so it's
the one process we're entitled to replace. The data dir is never touched, so
connections / secrets / the bearer survive.

- **`service install`** plans via a pure, unit-tested `planServiceInstall`:
  - `noop` — supervised service already runs this version
  - `reinstall` — supervised service runs an older/unreachable version → bootout→bootstrap repoints + restarts
  - `takeover-then-install` — something else holds the data dir → SIGTERM it, wait until it stops serving, clear its records, then install
- **Supervised daemon startup** (`runDaemonSession` under `EXECUTOR_SUPERVISED`)
  now takes over a surviving predecessor (manifest owner *and* a pointer-recorded
  `daemon run`) instead of failing and crash-looping under `KeepAlive`.

## Proof

`e2e/scripts/repro-upgrade-takeover.ts` (real macOS VM, real launchd):

| | result |
|---|---|
| before | `install exit=1`; predecessor (pid 631) keeps `:4789`; **NO-TAKEOVER** |
| after  | `install exit=0`; "Replaced the running Executor cli-daemon (pid 635)"; supervised service owns `:4789` (new pid); exactly one daemon — **TOOK-OVER** |

Plus unit tests for `planServiceInstall` (noop / reinstall / takeover / fresh).

### DIFF
--- a/apps/cli/src/daemon.test.ts
+++ b/apps/cli/src/daemon.test.ts
@@ -7,6 +7,7 @@ import {
   canAutoStartLocalDaemonForHost,
   isDevCliEntrypoint,
   isExecutorServerReachable,
+  planServiceInstall,
 } from "./daemon";
 
 describe("isDevCliEntrypoint", () => {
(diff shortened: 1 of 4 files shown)


RhysSullivan:
Superseded by the daemon-ownership rework that landed in #1175. `service install` now takes over a running daemon instead of refusing: `apps/cli/src/main.ts` has `takeOverActiveLocalServer` and a `takeover-then-install` install plan wired into the `service install` flow. Closing as done.

PR #1029, opened 2026-06-15
Stop reporting clean sidecar shutdowns as desktop crashes

RhysSullivan:
## Problem

The desktop app filed spurious crash reports of the form:

> Sidecar exited unexpectedly (code=130 signal=null)

`code=130` is `128 + SIGINT(2)` — the sidecar was **interrupted by SIGINT and shut down cleanly through its own signal handler**. The attached `stderrTail` was empty, confirming no fault occurred.

It got misclassified because `sidecar.ts` only treats a post-boot exit as *expected* when the desktop itself called `stopSidecar` (which sends **SIGTERM**). A SIGINT arriving any other way — most plausibly a process-group signal racing ahead of teardown, since the child shares the Electron main process's group — fell through to `reportSidecarCrash` **and** swapped the live UI for the crash screen, over what was really a healthy shutdown.

## Fix

Classify the post-boot exit on its **exit code/signal alone**:

1. **Quit-intent latch** — `markAppQuitting()` is set synchronously at the top of `before-quit`, *before* `stopConnection`'s `await`. Once the app is quitting, no sidecar exit is reported, closing the race where a SIGINT beats the SIGTERM.
2. **Clean-shutdown classification** — a graceful or signal-driven exit (code `0`/`130`/`143`, or a raw `SIGINT`/`SIGTERM`) is a shutdown, not a crash: it still surfaces the recovery screen (the web UI is dead, the user needs a way out) but is **not** reported. Abnormal exits — other non-zero codes, `SIGKILL`, a `null` code — still report exactly as before.

stderr is deliberately **not** consulted: the buffer is a rolling session tail (a startup log line makes it non-empty), so it's a poor proxy for "errored at exit". A genuine internal fault in the daemon is reported with a real stack by the sidecar's own crash reporter (`@sentry/bun`); this main-process path is the coarse "the child vanished" net, so it should fire only on abnormal exits.

The decision logic lives in a pure, electron-free helper (`sidecar-exit.ts`).

## Proof — black-box, through the real app

The on-screen flow can't show this bug: a clean shutdown and a hard kill paint the **same** recovery screen — the only difference is what reaches Sentry. So [`e2e/desktop/sidecar-clean-shutdown-not-reported.test.ts`](../blob/claude/sidecar-shutdown-false-crash/e2e/desktop/sidecar-clean-shutdown-not-reported.test.ts) launches the real Electron app pointed at a **local Sentry envelope sink** (via a non-packaged `EXECUTOR_DESKTOP_SENTRY_DSN` seam in `diagnostics.ts`, dead in production) and asserts:

- **SIGKILL** → a `Sidecar exited unexpectedly` crash **is** reported, and
- **SIGINT** → the recovery screen shows but **nothing** is reported,

with a **second SIGKILL as an ordering fence** — Sentry's transport is FIFO per client, so once the fence's envelope has arrived, any envelope the SIGINT would have produced is already present too; its absence is real, not just slow.

Verification:
- `vitest run --project desktop sidecar-clean-shutdown-not-reported` → **1 passed** (~25s)
- existing `sidecar-crash-screen` scenario (SIGKILL → crash screen → restart heals) → **still passes** (~16s), confirming SIGKILL is still classified as a crash
- `apps/desktop` + `e2e` typecheck, repo `oxlint`, `oxfmt --check` — all clean

## Notes for reviewers

- **Test seam:** `EXECUTOR_DESKTOP_SENTRY_DSN` overrides the baked-in DSN **only when `!app.isPackaged`**. Production is always packaged, so it can never redirect a real user's reports.
- **Coverage boundary:** the e2e exercises the dev sidecar (exit `0` on SIGINT) and SIGKILL. The prod-only `130`/`143` codes aren't hit by the dev-electron target; they're covered by the pure classifier's construction. The earlier pure unit test was dropped in favour of this black-box proof.

### DIFF
--- a/apps/desktop/src/main/diagnostics.ts
+++ b/apps/desktop/src/main/diagnostics.ts
@@ -25,7 +25,13 @@ import log from "electron-log/main.js";
 import * as Sentry from "@sentry/electron/main";
 import { getServerSettings } from "./settings";
 
-const sentryDsn = __EXECUTOR_SENTRY_DSN__;
+// Packaged builds always use the DSN baked in at build time. The non-packaged
+// override is a test seam: the desktop e2e points crash reporting at a local
+// sink so it can assert what is (and isn't) reported. It can never redirect a
+// real user's reports — production is always packaged, where the override is
+// dead and the baked DSN wins.
+const sentryDsn =
+  (app.isPackaged ? undefined : process.env.EXECUTOR_DESKTOP_SENTRY_DSN) || __EXECUTOR_SENTRY_DSN__;
 
 // The informal cross-tool opt-out (consoledonottrack.com). Checked before
 // any SDK initializes, and it covers all three processes because the
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -23,6 +23,7 @@ import {
   startSidecar,
   stopSidecar,
   onUnexpectedSidecarExit,
+  markAppQuitting,
   SidecarPortInUseError,
   type SidecarConnection,
 } from "./sidecar";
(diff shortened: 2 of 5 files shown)


RhysSullivan:

PR #1031, opened 2026-06-15
Toolkits

RhysSullivan:
A **toolkit** is a named slice of an account's connections — each connection set to `off` / `read` / `full` — with its own MCP endpoint. An agent that connects with `/mcp?toolkit=<slug>` sees and can run **only** that slice; everything outside it is invisible and blocked. Toolkits come in two scopes: **workspace** (org-owned connections, shared with every member) and **personal** (org + the caller's own, private to them).

Shipped end-to-end as a self-contained plugin (`@executor-js/plugin-toolkits`) plus the small core seams it needs.

## What's included

- **Data + CRUD** — owner-scoped toolkit storage (workspace vs personal) and a typed CRUD API.
- **MCP narrowing** — connecting with a `?toolkit=` selector scopes the engine to that toolkit's slice across every surface: tool listing, search, descriptions, the connection/integration catalog, the built-in core tools, **and** execute (a guessed out-of-slice address is blocked, not just hidden).
- **Access modes** — `read` exposes only read-only tools; `full` exposes all; `off` removes the connection. Wildcard (`*`) entries track every account of an integration, including ones added later.
- **Per-toolkit policies** — `block` / `require_approval` rules compose with org-level policies via the existing stricter-wins lattice. A toolkit can only *tighten* — org guardrails always apply.
- **Console UI** — a Toolkits page (sidebar nav on every host) with a list split by scope and a per-toolkit editor, **URL-routed** (`/plugins/toolkits/`, `/plugins/toolkits/<id>`, `/plugins/toolkits/new/<scope>`) — deep-linkable, with working browser back/forward.
- **Cross-environment** — works on self-host, local, and the cloud/Cloudflare Durable-Object MCP path.

## Architecture notes

- **Core stays vocabulary-neutral.** The feature is a plugin; core knows nothing about "toolkits." Narrowing is expressed through a generic **`RequestScope`** capability — a plugin contributes *data* (`decideTool` / `allowsConnection` / `allowsIntegration` / `inheritOrgPolicies`), and **core owns enforcement centrally** across all surfaces. This replaced an earlier executor-wrapper approach that re-implemented narrowing per-surface and was fail-open on any surface it didn't override (including the built-in `connections.list` / `integrations.list` core tools, which could otherwise enumerate the full account). Unknown selectors fail closed to an empty scope; mutation/admin surfaces are default-denied under a scope.
- **Plugin pages can route by URL.** The plugin page host now exposes the URL subpath + a navigate helper to plugin pages (`usePluginRoute` / `usePluginNavigate`), so any plugin can do internal URL routing — no change to the shared route contract or per-host wiring.

## Testing

13 self-host e2e files / 16 tests covering CRUD + scope visibility, cross-user isolation, read-mode (GET vs write), per-toolkit and inherit-org policies, wildcard auto-include, fail-closed selectors, the core-tool scope boundary (scoped `connections.list` / `integrations.list` return only in-slice; scoped sessions can't create connections or policies), and the console UI flow (create, deep-link, browser back). Full `typecheck` / `lint` / `format:check` green.

## Preview

Console: https://executor-preview-pr-1031.executor-e2e.workers.dev · MCP: `https://executor-preview-pr-1031.executor-e2e.workers.dev/mcp`

### DIFF
--- a/apps/cloud/executor.config.ts
+++ b/apps/cloud/executor.config.ts
@@ -3,6 +3,7 @@ import { openApiHttpPlugin } from "@executor-js/plugin-openapi/api";
 import { mcpHttpPlugin } from "@executor-js/plugin-mcp/api";
 import { graphqlHttpPlugin } from "@executor-js/plugin-graphql/api";
 import { workosVaultPlugin, type WorkOSVaultClient } from "@executor-js/plugin-workos-vault";
+import { toolkitsPlugin } from "@executor-js/plugin-toolkits/server";
 
 // ---------------------------------------------------------------------------
 // Single source of truth for the cloud app's plugin list.
@@ -52,5 +53,6 @@ export default defineExecutorConfig({
         credentials: workosCredentials ?? { apiKey: "", clientId: "" },
         ...(workosVaultClient ? { client: workosVaultClient } : {}),
       }),
+      toolkitsPlugin(),
     ] as const,
 });
--- a/apps/cloud/package.json
+++ b/apps/cloud/package.json
@@ -44,6 +44,7 @@
     "@executor-js/plugin-graphql": "workspace:*",
     "@executor-js/plugin-mcp": "workspace:*",
     "@executor-js/plugin-openapi": "workspace:*",
+    "@executor-js/plugin-toolkits": "workspace:*",
     "@executor-js/plugin-workos-vault": "workspace:*",
     "@executor-js/react": "workspace:*",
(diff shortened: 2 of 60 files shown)


RhysSullivan:
Superseded by a reviewable stack with the same final diff, split into logical layers (e2e last):
- #1036 — plugin extensibility seams (RequestScope + plugin-page routing)
- #1037 — toolkits plugin
- #1038 — mount on every host
- #1039 — e2e coverage

PR #1033, opened 2026-06-15
Make executor mcp attach to daemon

RhysSullivan:
## Summary
- Make `executor mcp` attach to the active local daemon.
- Add local e2e coverage.

## Tests
- `bun --bun vitest run --project local local/cli-mcp-daemon-attach.test.ts`
- `bun --bun vitest run src/stdio-integration.test.ts`


RhysSullivan:
Superseded by #1196. That PR takes a stronger approach: instead of the first `executor mcp` process starting a server in-process and bridging to itself (which tied the shared owner's lifetime to a transient client), it ensures a durable detached daemon via the merged start-lock primitive and bridges to that, so no client ever owns the database. The cold-start election is verified on macOS, Linux, and Windows.

PR #1052, opened 2026-06-19
better svgs for openai, spotify & sentry

maria-rcks:
| before | after |
|-------|-------|
| image | image |


RhysSullivan:
bless 🙏

PR #1059, opened 2026-06-20
mcp: surface auth-gated failures

gjermundgaraba:
I noticed after trying to add the official excalidraw mcp server that a basic 401 error (non-compliant mcp error (no `Bearer WWW-Authenticate`, no RFC 9728 metadata)), would not let me add the server at all and progress to the steps needed to configure the auth headers.

The fix applied here attempts to solve this by handling non-mcp-spec compliant auth failures the same way as compliant ones.

If you don't want to handle non-mcp-spec compliant auth failures like this, feel free to disregard the PR.

Additional note, during testing and review of the changes, I also found another issue that is folded in here for tool invocation: call-time 401/403 responses (and OAuth reauthorization) are now classified as actionable auth failures ("re-authenticate / update the connection") rather than generic tool errors, while ensuring raw upstream error bodies are never leaked into user-facing messages.


RhysSullivan:
thanks!

PR #1062, opened 2026-06-21
[codex] Handle OpenAPI file outputs in MCP

RhysSullivan:
## Summary

- Add a `ToolFile` value type and schema for tool outputs that are files.
- Detect OpenAPI `format: byte` and `format: binary` responses, normalize bytes into `ToolFile`, and preserve normal error bodies for non-2xx responses.
- Project `ToolFile` results in the MCP host as model-visible image/text/resource content while stripping base64 bytes from structured metadata.

## Why

OpenAPI byte fields currently surface as ordinary JSON strings, so agents and clients end up seeing or manipulating base64 directly. This makes binary outputs, especially image attachments, hard for models to consume correctly.

## Validation

- `bunx vitest run src/tool-server.test.ts` in `packages/hosts/mcp`
- `bunx vitest run src/sdk/non-json-body.test.ts` in `packages/plugins/openapi`
- `bunx vitest run src/tool-result.test.ts` in `packages/core/sdk`
- Package typechecks for `packages/hosts/mcp`, `packages/plugins/openapi`, `packages/core/sdk`, and `packages/core/execution`
- `bunx oxfmt --check ...`
- `git diff --check`
- Deployed worker smoke test: live MCP attachment calls rendered image content blocks instead of base64 text

### DIFF
--- a/packages/core/sdk/src/tool-result.ts
+++ b/packages/core/sdk/src/tool-result.ts
@@ -34,52 +34,22 @@ export const ToolFileSchema = Schema.TaggedStruct("ToolFile", {
   name: Schema.optional(Schema.String),
   mimeType: Schema.String,
   encoding: Schema.Literal("base64"),
-  data: Schema.String,
-  byteLength: Schema.Number,
+  data: Schema.String.annotate({
+    description: "Base64-encoded file bytes.",
+    contentEncoding: "base64",
+  }),
+  byteLength: Schema.Int.annotate({
+    description: "Raw file size in bytes before base64 encoding.",
+  }),
 });
 
 export type ToolFile = typeof ToolFileSchema.Type;
 
-export const ToolFileJsonSchema = {
-  type: "object",
-  properties: {
-    _tag: { const: "ToolFile" },
-    name: { type: "string" },
-    mimeType: { type: "string" },
-    encoding: { const: "base64" },
-    data: {
-      type: "string",
-      contentEncoding: "base64",
-      description: "Base64-encoded file bytes.",
-    },
-    byteLength: {
-      type: "integer",
-      description: "Raw file size in bytes before base64 encoding.",
-    },
-  },
-  required: ["_tag", "mimeType", "encoding", "data", "byteLength"],
-  additionalProperties: false,
-} as const;
+export const ToolFileJsonSchema = Schema.toJsonSchemaDocument(ToolFileSchema).schema;
 
-export const ToolFile = {
-  make: (input: {
-    readonly name?: string;
-    readonly mimeType: string;
-    readonly data: string;
-    readonly byteLength: number;
-  }): ToolFile => ({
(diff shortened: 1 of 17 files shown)


packages/core/sdk/src/tool-result.ts
```diff
@@ -30,6 +30,57 @@ export const ToolHttpMetaSchema = Schema.Struct({
...
+      description: "Base64-encoded file bytes.",
+    },
+    byteLength: {
+      type: "integer",
+      description: "Raw file size in bytes before base64 encoding.",
+    },
+  },
+  required: ["_tag", "mimeType", "encoding", "data", "byteLength"],
+  additionalProperties: false,
+} as const;
```
RhysSullivan: why are we manually defining json schema cant you use effects json schema

RhysSullivan: i miss fable

packages/core/sdk/src/tool-result.ts
```diff
@@ -30,6 +30,57 @@ export const ToolHttpMetaSchema = Schema.Struct({
...
+    readonly data: string;
+    readonly byteLength: number;
+  }): ToolFile => ({
+    _tag: "ToolFile",
+    ...(input.name ? { name: input.name } : {}),
+    mimeType: input.mimeType,
+    encoding: "base64",
+    data: input.data,
+    byteLength: input.byteLength,
+  }),
```
RhysSullivan: this manual function is ugly

packages/core/sdk/src/tool-result.ts
```diff
@@ -30,6 +30,57 @@ export const ToolHttpMetaSchema = Schema.Struct({
...
+} as const;
+
+const isUnknownToolFile = Schema.is(ToolFileSchema);
+
+export const isToolFile = (value: unknown): value is ToolFile => isUnknownToolFile(value);
+
```
RhysSullivan: when is it ever gonna be unknown

packages/hosts/mcp/src/tool-server.ts
```diff
@@ -263,17 +266,204 @@ const formatBoundaryError = (err: unknown): { name?: string; message: string; st
...
-const toMcpResult = (formatted: ReturnType<typeof formatExecuteResult>): McpToolResult => ({
-  content: [{ type: "text", text: formatted.text }],
-  structuredContent: formatted.structured,
-  isError: formatted.isError || undefined,
+type FormattedExecuteInput = Parameters<typeof formatExecuteResult>[0];
+
+const TEXT_FILE_CONTENT_MAX_CHARS = 64_000;
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === "object" && value !== null && !Array.isArray(value);
```
RhysSullivan, line 279: i miss fable

PR #1096, opened 2026-06-23
Mark Google media downloads as binary responses

batumilove:
## Summary
- Mark Google Discovery methods with media-download support and no JSON response schema as binary OpenAPI responses.
- This lets the existing OpenAPI runtime return ToolFile base64 payloads instead of UTF-8 JSON strings for Drive files.export.
- Add coverage for Drive files.export response conversion.

## Local verification
- git diff --check passed.
- Full targeted Vitest is pending local dependency installation. bun install --frozen-lockfile is still resolving on the deployment host.

## Operational context
This fixes corruption where Drive binary exports such as PDF/DOCX were serialized through JSON strings and non-UTF-8 bytes became U+FFFD replacement characters.

### DIFF
--- a/packages/plugins/google/src/sdk/discovery.test.ts
+++ b/packages/plugins/google/src/sdk/discovery.test.ts
@@ -247,6 +247,64 @@ it.effect("converts Google Discovery documents into Executor-preserving OpenAPI 
   }),
 );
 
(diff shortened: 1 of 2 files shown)


RhysSullivan:
thanks!

PR #1097, opened 2026-06-23
feat(execution): add execution observer foundation

aryasaatvik:
## Summary

Add an Effect-native execution lifecycle observer that plugins can use for durable history, metrics, indexing, or cache maintenance without coupling those products to the execution engine. The hook complements existing spans and OpenTelemetry rather than replacing them.

## Contract

```ts
type ExecutionEvent =
  | ExecutionStarted
  | ToolCallStarted
  | ToolCallFinished
  | InteractionStarted
  | InteractionResolved
  | ExecutionFinished;

definePlugin(() => ({
  runtime: {
    executionObserver: (extension) => ({ handle: (event) => Effect.void }),
  },
}));
```

```text
execute / executeWithPause / resume
  -> install the composed observer in Effect context
  -> emit execution, tool-call, and interaction events
  -> dispatch sequentially to plugin observers
  -> isolate and log non-interrupt observer failures
```

- Observer context survives inline and detached pause/resume fibers.
- Interrupts remain cancellations; ordinary observer failures cannot fail execution.
- Shared API and local execution stacks use the same plugin hook.
- Owner metadata preserves execution attribution.

## Contract update

The observer contract now carries what the engine already knew:

- `ExecutionFinished.output` — everything the code sent through `emit()`, in
  order, so an emit-only run records a result instead of nothing. The new
  `ExecutionOutputItem` mirrors the sandbox shape structurally, keeping the sdk
  free of the kernel package.
- `ExecutionStatus` gains `interrupted`, emitted from an `Effect.onExit`
  finalizer on both execute paths. The finalizer runs uninterruptibly, so a run
  torn down from outside (client abort, host backstop, sandbox shutdown) still
  closes in every observer instead of dangling as "running"; an interrupt-only
  cause maps to `interrupted` rather than `failed`.
- Failed `ToolResult` envelopes are reported as failed tool calls. Expected tool
  failures ride the success channel as `ToolResult.fail`, so `ToolCallFinished`
  now reports `failed` with `code: message` and keeps the envelope attached for
  inspection, instead of observing every upstream 4xx as a completed call.

## Validation

- Focused SDK observer tests: 5 passed.
- Engine lifecycle tests: 3 passed across inline and pause/resume execution.
- SDK, execution, API, and local package suites and typechecks passed.
- Format, lint, and all 36 applicable GitHub CI checks passed at `9f9373469`.

### Execution history delivery map

Prerequisites:

- [ ] **#1097 — Execution observer foundation** 👈 current
- [ ] #1119 — Aggregate and keyset queries

Follow-up PR-sized diffs:

- [Core persistence and read API](https://github.com/aryasaatvik/executor/compare/contrib/execution-history-integration-base...contrib/execution-history-plugin)
  - [Runs console UI](https://github.com/aryasaatvik/executor/compare/contrib/execution-history-plugin...contrib/execution-history-runs-ui)
    - [Cloudflare service-token actors](https://github.com/aryasaatvik/executor/compare/contrib/execution-history-runs-ui...contrib/execution-history-cloudflare-actors)

The fork also uses this observer for execution metrics, which remains a separate consumer. [View the complete fork comparison](https://github.com/UsefulSoftwareCo/executor/compare/main...aryasaatvik:executor:dev).

https://claude.ai/code/session_016tQXmEycQ2gmmW7LJt6Nhz


aryasaatvik:
Rebased onto current main. The conflicts were with #1919's pause/resume rework in `engine.ts` and the `orgWrites` and span additions in `execution-stack.ts`. `resumeExecution` is byte-identical to main: `InteractionResolved` is emitted from the sandbox fiber when the response deferred settles, so the new `pendingResumes` shape and the org-write-access swaps needed no observer edits. All seven commits are preserved.

https://claude.ai/code/session_016tQXmEycQ2gmmW7LJt6Nhz

aryasaatvik:
Head moved `a3be7ac7b` → `7decefbca`: one commit added on top, `feat(sdk,execution): carry output, interruption, and envelope failures to observers`.

It extends the observer contract with `ExecutionFinished.output` (everything the code sent through `emit()`), an `interrupted` execution status emitted from an `Effect.onExit` finalizer so an externally torn-down run still closes, and failed `ToolResult` envelopes surfaced as failed tool calls with the envelope kept for inspection. Three tests were added to `engine-observer.test.ts`.

Verified on the new head: `packages/core/execution` 79/79 and `packages/core/sdk` 897/897 vitest passing; typecheck clean for `@executor-js/sdk`, `@executor-js/execution`, `@executor-js/host-mcp`; format and lint clean.

https://claude.ai/code/session_016tQXmEycQ2gmmW7LJt6Nhz

PR #1101, opened 2026-06-23
Route OAuth token HTTP through hosted egress guard

RhysSullivan:
## What changed

- Add a shared hosted Fetch API adapter for libraries that require fetch, using the same hosted outbound URL guard as the hosted HttpClient layer.
- Pass that adapter into `createExecutor` as the generic `fetch` boundary while normal SDK and plugin HTTP stays on `httpClientLayer`.
- Route OAuth token exchange, client-credentials, and refresh requests through that guarded fetch boundary.

## Why

OAuth4WebAPI requires a fetch-compatible hook for token endpoint calls. Hosted executors need that forced boundary to use the same private-network guard as the rest of hosted outbound HTTP.

## Validation

- `bun --bun vitest run src/oauth-helpers.test.ts src/oauth-flow.test.ts src/hosted-http-client.test.ts` from `packages/core/sdk`
- `bun --bun run typecheck` from `packages/core/sdk`
- `bun --bun run typecheck` from `packages/core/api`

## Stack

Base: `fix/hosted-egress-dns-guard`

Previous: #1100

Next: `fix/mcp-hosted-egress-guard`

### DIFF
--- a/packages/core/api/src/server/scoped-executor.ts
+++ b/packages/core/api/src/server/scoped-executor.ts
@@ -41,7 +41,7 @@ import {
   type Executor,
   type StorageFailure,
 } from "@executor-js/sdk";
-import { makeHostedHttpClientLayer } from "@executor-js/sdk/host-internal";
+import { makeHostedFetch, makeHostedHttpClientLayer } from "@executor-js/sdk/host-internal";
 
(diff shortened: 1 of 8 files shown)


RhysSullivan:
Superseded by batch merge #1106.

PR #1108, opened 2026-06-23
Connection health checks (liveness) with OpenAPI backing

RhysSullivan:
Base of a 4-PR stack. A connection declares ONE authenticated operation that answers "is this credential still alive?". The probe maps HTTP status to a health state (2xx healthy, 401/403 expired, other degraded, unset unknown) — built for short-lived OAuth tokens (Google's 7-day dev-token revocation).

- Core vocabulary (HealthStatus, HealthCheckSpec, HealthCheckResult, HealthCheckCandidate, classifyHttpStatus, ranking) + dispatch (`integrations.healthCheck.{get,candidates,set}`, `connections.{checkHealth,validate}`).
- OpenAPI backing: ranked candidates, persist the chosen spec (merging over raw config so provider supersets keep their keys), run the probe.
- React: per-connection status dot + "Check now"; a self-hiding operation editor; and **"Check the key works" in the Add Connection modal** — probe the pasted key before saving, and when no check is configured yet, pick a read-only operation inline (a healthy probe saves it as the integration's check).

The operation picker is a base-ui combobox whose popup portals out of the surrounding Radix modal, so this PR also carries the fixes that make it usable in a modal (folded in from the former combobox-sheet PR, since both the editor sheet and the Add Connection dialog need them): pointer-events on the popup, an outside-interaction guard on the dialog AND the sheet so clicking an option does not dismiss the modal, and a non-modal editor sheet so the popup list can scroll.

Built on the self-contained add-connection modal (#1126): the key-check state lives inside the modal view, so closing unmounts and resets it (no hand-reset).

e2e: saved connection healthy -> expired; unconfigured -> unknown; the operation picker filters a hundreds-long list on the edit sheet and add screen; Add Connection checks the key against an inline-picked operation BY MOUSE and persists it; the editor-sheet combobox is clickable and scrollable.

### DIFF
--- a/e2e/scenarios/health-checks-ui.test.ts
+++ b/e2e/scenarios/health-checks-ui.test.ts
@@ -1,0 +1,573 @@
+// Cross-target (browser): the UI side of connection health checks, the feature
+// that answers "has this credential expired?" (the Google 7-day dev-token case).
+// These scenarios pin the operation picker, the part that lets a user choose
+// WHICH call the probe runs:
+//
+//  1. Add Connection: with no health check configured yet, the user picks a
+//     read-only operation inline, clicks "Check the key works", the probe runs,
+//     and the picked operation is saved as the integration's health check.
+//  2. Edit sheet, large spec: typing into the operation combobox filters a
+//     hundreds-long candidate list down to the one match, and committing it
+//     stores the real operation (not the freeform text typed to find it).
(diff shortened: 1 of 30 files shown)


RhysSullivan:

RhysSullivan:
Replacement: #1251

PR #1109, opened 2026-06-23
Health checks for Microsoft Graph and Google

RhysSullivan:
Extends connection health checks to the provider plugins.

- **Microsoft Graph** and **Google** wire the four OpenAPI health-check hooks to the shared backing (same store + config superset). Google auto-configures the People API identity check when the bundle includes it; Graph exposes the picker so the user selects the identity operation.
- **MCP** gets a liveness-only probe (connect + list tools, classify the result). MCP has no usable identity source, so no identity is derived; the connection keeps its user-given label.
- The MCP HTTP-status extractor moved to a small shared module so the connect path can surface a 401/403 to the liveness check.

E2E: `health-checks-providers.test.ts` (MCP healthy→expired + validate; Google auto-default + candidate response fields).

Stacked on #1108.

1. #1110
2. #1108
3. #1111
4. #1112
5. **#1109** 👈 current

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -342,6 +342,7 @@
         "@executor-js/api": "workspace:*",
         "@executor-js/emulate": "^0.7.5",
         "@executor-js/mcporter": "^0.11.4",
+        "@executor-js/plugin-google": "workspace:*",
         "@executor-js/plugin-graphql": "workspace:*",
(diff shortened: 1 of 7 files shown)


RhysSullivan:

RhysSullivan:
Replacement: #1251

PR #1110, opened 2026-06-23
Fix combobox inside the health-check editor sheet

RhysSullivan:
The operation/identity comboboxes portal their popup to `<body>`, outside the sheet's dialog subtree, so a modal Radix sheet froze them: body `pointer-events` were disabled (couldn't click/select) and `react-remove-scroll` locked wheel scroll.

- The editor sheet is non-modal, which drops both locks while keeping the dim overlay.
- The popup gets `pointer-events: auto` (defensive for any modal-dialog combobox).
- The sheet stays open when an interaction originates inside a combobox popup, so clicking an option no longer dismisses it.

Adds e2e guards that click and wheel-scroll the popup inside the sheet.

Stacked on #1109.

1. #1108
2. #1111
3. #1112
4. #1109
5. **#1110** 👈 current

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -342,6 +342,7 @@
         "@executor-js/api": "workspace:*",
         "@executor-js/emulate": "^0.7.5",
         "@executor-js/mcporter": "^0.11.4",
+        "@executor-js/plugin-google": "workspace:*",
         "@executor-js/plugin-graphql": "workspace:*",
         "@executor-js/plugin-mcp": "workspace:*",
         "@executor-js/plugin-microsoft": "workspace:*",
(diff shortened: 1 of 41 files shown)


RhysSullivan:
Folded into the base PR (#1108). The combobox-in-modal fixes (popup pointer-events, dialog + sheet outside-interaction guards, non-modal editor sheet) are prerequisites for the base PR's operation picker in both the editor sheet and the Add Connection modal, so they belong in #1108 rather than as a trailing PR. The scroll/click e2e moved to #1108 (sheet) and the identity-combobox click into #1111.

PR #1111, opened 2026-06-23
Connection account info: derive identity from the health-check probe

RhysSullivan:
Stacked on #1108. Layers account identity on top of the liveness probe: a health check can name a response field whose value identifies the connected account, so the same probe answers "is this alive?" and "whose account is this?". Adds the typed identity-field picker, the live preview (probe a pasted key, see the real response), the account-row identity label, and the key-first connect modal (validate then auto-name). Also restores the OAuth-client deep-link prefill the key-first reorder had dropped.

1. #1110
2. #1108
3. **#1111** 👈 current
4. #1112
5. #1109


RhysSullivan:

RhysSullivan:
Replacement: #1251

PR #1112, opened 2026-06-23
MCP connection liveness health checks

RhysSullivan:
Stacked on #1111. Gives MCP connections a liveness probe (dial the server, list tools, classify 401/403 as expired). MCP is liveness-only: no usable identity source, so no identity is derived and the operation/identity editor stays hidden; the generic status dot + "Check now" still render. Factors HTTP-status extraction into a shared helper so the probe can classify auth failures.

1. #1110
2. #1108
3. #1111
4. **#1112** 👈 current
5. #1109


RhysSullivan:

RhysSullivan:
Replacement: #1251

PR #1114, opened 2026-06-24
Add Vercel eve agent adapter (@executor-js/host-eve)

jackulau:
Expose Executor's codemode surface to a Vercel eve agent as two tools, execute and resume, mirroring the MCP host. Backed by the promise execution engine, so the package has no eve runtime dependency: the factory returns plain objects shaped for eve's defineTool.

Includes unit tests plus integration tests that drive the adapter against the real execution engine and QuickJS sandbox.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -662,6 +662,24 @@
         "vitest": "catalog:",
       },
     },
+    "packages/hosts/eve": {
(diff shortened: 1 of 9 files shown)


jackulau:
wip - taking feedback.

PR #1121, opened 2026-06-24
Show an actionable message when MCP OAuth dynamic registration is rejected

### LINKED ISSUE #770: Vercel MCP OAuth setup fails with invalid_redirect_uri
## Summary

Adding the Vercel MCP source at `https://mcp.vercel.com` fails during OAuth setup. The UI reports that dynamic authorization setup failed because the redirect URI is not approved by the authorization server.

## Steps to reproduce

1. Open **Add MCP Source**.
2. Select or configure `mcp.vercel.com` with server URL `https://mcp.vercel.com`.
3. Use OAuth authentication.
4. Try to add/connect the source.

## Actual result

The source cannot be added. The UI shows:

```text
Dynamic authorization setup failed: Dynamic Client Registration failed: invalid_redirect_uri — The provided redirect URIs are not approved for use by this authorization server.
```

The form also shows `OAuth required to discover tools`.

## Expected result

The OAuth flow should complete successfully, or the app should provide a supported redirect URI/setup path for Vercel MCP.

## Notes

Observed from the Add MCP Source screen while configuring:

- Display name: `Vercel MCP`
- Namespace: `vercel_mcp`
- Server URL: `https://mcp.vercel.com`
- Auth: OAuth
- Connection saved to: Personal

Screenshot was provided with the report.

Image

jackulau:
Fixes #770.

## Problem

Adding the Vercel MCP source (`https://mcp.vercel.com`) with OAuth fails during automatic setup with:

```
Dynamic Client Registration failed: invalid_redirect_uri - The provided redirect URIs are not approved for use by this authorization server.
```

## Root cause

Transparent Dynamic Client Registration (DCR) registers Executor's browser origin (`window.location.origin + /api/oauth/callback`) as the client `redirect_uris`. Vercel's authorization server only approves loopback redirect URIs (`http://localhost` / `http://127.0.0.1`, any port) for anonymous DCR, a deliberate RFC 8252 stance. Any non-loopback origin (a hosted https deployment, a tailnet hostname, a LAN IP) is rejected with `invalid_redirect_uri`. Verified against the live registration endpoint: loopback URIs register successfully; `https://localhost`, `*.vercel.app`, tailnet, and arbitrary https all return `invalid_redirect_uri`.

This is not unique to Vercel; it affects any authorization server that enforces loopback-only DCR, so the fix is carrier-agnostic.

## Fix

- **SDK**: when DCR fails with `invalid_redirect_uri` and the redirect URI Executor used is non-loopback, the registration error now carries an actionable message that names the loopback-only requirement, the redirect URI in use, and the two recovery paths (register an OAuth app manually with that redirect URL approved, or run Executor on `http://localhost`). The generic message is preserved for every other failure, and the loopback hint is gated on the URI actually being non-loopback so a localhost user is never told to use localhost.
- **React**: the transparent DCR connect flow previously swallowed the registration error and showed a generic "register an app" toast. It now threads the server's failure message into the bring-your-own-app fallback and surfaces it, so the user sees why automatic setup was declined while still landing on the manual-registration recovery path.

## Testing

- `oauth-register-dynamic.test.ts`: added cases for a non-loopback redirect URI (actionable hint shown) and a loopback redirect URI still rejected (generic message kept, no false hint). The OAuth test server gains an `approveRedirectUri` option to mimic a loopback-only allowlist.
- `add-account-modal.test.ts`: added a case asserting the registration failure message reaches the fallback outcome and the OAuth start is skipped.
- `typecheck`, `format:check`, and `oxlint --deny-warnings` all pass.

## Coverage note

There is no live end-to-end test against the real Vercel authorization server (anonymous DCR would pollute it and requires a hosted origin). The loopback-only rejection is reproduced in unit tests via the test server mirroring the verified live behavior.


RhysSullivan:
Updated the DCR failure state so the OAuth section stays visible while the inline error card handles the rejection.

PR 1121 DCR failure state

jackulau:
> Updated the DCR failure state so the OAuth section stays visible while the inline error card handles the rejection.
> 
> PR 1121 DCR failure state [ PR 1121 DCR failure state ](https://camo.githubusercontent.com/a0bde8b85f71869a0723396c92b6a58368f4acac61940f35bced9a117b18c66c/687474703a2f2f3130302e38312e3231392e34353a34393232312f70722d313132312d6463722d72656a656374696f6e2e676966) [ ](https://camo.githubusercontent.com/a0bde8b85f71869a0723396c92b6a58368f4acac61940f35bced9a117b18c66c/687474703a2f2f3130302e38312e3231392e34353a34393232312f70722d313132312d6463722d72656a656374696f6e2e676966)

I'm getting a error trying to view the media content

PR #1131, opened 2026-06-25
perf: reduce /api/tools server time on the read path

amondnet:
## Summary

Sub-task of #3 (Fix A): cut `/api/tools` server time, the measured long pole on the web UI.

Both issues called out in #4 were verified against the code first, then fixed TDD-style.

### 1. Skip the stale-sync scan on reads (`packages/core/sdk`)

`tools.list` ran `syncStaleConnectionTools` unconditionally: a read of revised integrations plus a connection scan, on every call. Because `config_revised_at` stays set after the first config change, the steady state was two full reads that rebuilt nothing.

Now the read of revised integrations doubles as a **revision watermark**. Once every connection is confirmed synced at the current watermark, that is cached per binding and the connection scan is skipped until a new revision moves the watermark:

- The cache is a `WeakMap` keyed by the underlying DB handle, then by tenant+subject. In production every per-request scoped executor in an isolate shares one long-lived handle, so the cache survives executor rebuilds and the saving lands; each test builds a fresh DB object, so entries are naturally isolated (no cross-test contamination) and GC'd with the handle.
- The watermark is read fresh every call, so a revision on any isolate busts the cache on the next read. Lazy cross-subject convergence stays correct.
- A failed rebuild is left uncached so the next read retries it (best-effort behavior preserved).

Net: steady-state `tools.list` drops from two table reads to one.

### 2. Filter `listOperations` at the storage layer (`packages/plugins/openapi`)

`listOperations` read the whole `operation` collection and filtered by integration in memory. Even passing a key prefix did not reach the DB (the facade applied it in JS after a full scan).

- Added `keyPrefixes` to the plugin-storage list input; the facade now pushes `key starts with` clauses to the storage layer (`LIKE 'prefix%'`).
- The store passes both the v2 hashed prefix (`op.<hash>.`) and the legacy plaintext prefix (`<slug>.`), keeping the `data.integration` JS guard as the source of truth, so un-migrated legacy rows are never dropped and `LIKE` wildcard over-match is harmless.

## Tests

- New: facade `list` with `keyPrefixes` over real SQLite; openapi store `listOperations` (per-integration, legacy-safe, prefix-narrowing); `syncStaleConnectionTools` skip / cache-bust / retry, driven by a per-table query counter.
- Existing cross-subject convergence coverage (`updateSpec propagates to OTHER subjects' personal connections`) exercises the new cache + `starts with` path on real SQLite and stays green.
- `packages/core/sdk` 352/352, `packages/plugins/openapi` 151/151. `format:check`, `lint`, and typecheck clean.

Closes #4

### DIFF
--- a/.changeset/reduce-api-tools-server-time.md
+++ b/.changeset/reduce-api-tools-server-time.md
@@ -1,0 +1,6 @@
+---
+"@executor-js/sdk": patch
+"@executor-js/plugin-openapi": patch
+---
+
+Reduce `tools.list` server time. The read path now skips the stale-connection-tools sync once every connection is synced at the current revision watermark (cached per binding and busted by any new config revision), instead of scanning connections on every call. The openapi operation store also filters `listOperations` by integration at the storage layer (a key-prefix `LIKE` covering both the hashed and legacy key schemes) rather than reading the whole collection and filtering in memory.
--- a/packages/core/sdk/src/executor.test.ts
+++ b/packages/core/sdk/src/executor.test.ts
@@ -1,7 +1,9 @@
 import { describe, expect, it } from "@effect/vitest";
 import { Data, Effect, Predicate, Result } from "effect";
+import { withQueryContext } from "@executor-js/fumadb/query";
 
 import { ToolNotFoundError } from "./errors";
+import { StorageError } from "./fuma-runtime";
 import {
   AuthTemplateSlug,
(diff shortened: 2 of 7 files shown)


amondnet:
Re-opening against chatbot-pf/executor (intended base); opened here by mistake due to fork default.

PR #1136, opened 2026-06-26
feat(google): agent setup tools for Google integration

RhysSullivan:
## What

Adds agent-callable setup tools to the Google plugin so an agent can configure Google products through chat, mirroring the existing web product-picker UX. No generalized source-setup service: this is scoped to the Google plugin and reuses the existing extension and presets.

Three static-source tools:

- **listProducts** (read-only): catalog of available Google products with each product's OAuth audience tier and special-consent flags.
- **addBundle** (requires approval): resolves selected product ids to Discovery specs and registers a single integration covering them. Echoes back the products, surfaces audience warnings for products that need special consent, and returns the OAuth next steps.
- **setupStatus** (read-only): reports whether Google is currently configured.

## Secret handling

The tools never accept a client secret. Google has no Dynamic Client Registration, so the user brings their own Google Cloud OAuth client. Secret entry stays in the web UI via the existing OAuth handoff; the agent tools point the agent at that flow and explicitly avoid asking for the secret in chat. This keeps the client secret off the agent channel.

## Shared logic

The product catalog and audience-warning helpers are consolidated in the SDK presets module and reused by both the new agent tools and the React accounts panel (the panel previously held a local duplicate of the audience-warning map).

## Tests

New Effect Vitest suite drives all three tools through `executor.execute`, with a stubbed HTTP client keyed on the canonical Discovery URLs:

- listProducts returns the catalog and makes zero network calls
- addBundle resolves product ids to a single integration, echoes products, and includes the OAuth next steps
- unknown product id fails before any network call and names the bad id back to the caller
- empty selection fails with `no_products_selected` and no network call
- a mixed bundle surfaces exactly the expected audience warnings
- setupStatus transitions unconfigured to configured

Verification: `format:check`, `lint` (0 errors), `typecheck` (all packages), and the Google package test suite all pass.


RhysSullivan:
Superseded on main: #1366 moved Google/Microsoft onto the OpenAPI plugin, which replaces the approach here.

PR #1194, opened 2026-06-28
fix(cloud): stop PR preview builds failing on DO migrations (10211)

RhysSullivan:
## Problem

Preview (non-`main`) Workers Builds deploy via `wrangler versions upload`, which rejects any config containing a Durable Object migration with **error 10211** ("migrations must be applied via a non-versioned deployment"). Since `apps/cloud`'s `wrangler.jsonc` carries the DO migration ledger (`v1`/`v2`), every PR's preview build fails.

## Fix

`build.mjs` strips `migrations` from the generated `dist/server/wrangler.json` **only on non-`main` CI branches** (`WORKERS_CI=1` + `WORKERS_CI_BRANCH !== "main"`). Previews share production's already-applied DO state, so they don't need (and can't apply) migrations. `main` keeps migrations and applies them on the real, non-versioned deploy.

Fail-safe by construction: it only ever strips on a confirmed non-`main` CI branch, so it can't drop migrations from a production deploy. Confirmed via the CF build log that the strip takes effect (the earlier error moved from `10211` to a different one once stripped).

## Scope note

This intentionally does **not** delete the orphaned KV `McpSessionDO` class. That class is unbound and harmless, the migration ledger must stay regardless, and a delete-class PR's own preview build can't pass anyway. Left as-is with its one-line stub export.

## Verified locally
- typecheck 0, lint clean
- normal/`main` build keeps migrations; preview-branch build logs the strip and emits no migrations in the deploy config

### DIFF
--- a/apps/cloud/scripts/build.mjs
+++ b/apps/cloud/scripts/build.mjs
@@ -9,7 +9,7 @@
 
 import { spawnSync } from "node:child_process";
 import { randomBytes } from "node:crypto";
-import { rmSync } from "node:fs";
+import { readFileSync, rmSync, writeFileSync } from "node:fs";
 
 if (!process.env.VITE_PUBLIC_ANALYTICS_PATH) {
   process.env.VITE_PUBLIC_ANALYTICS_PATH = randomBytes(4).toString("hex");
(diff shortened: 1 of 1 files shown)


RhysSullivan:
Closing — unnecessary. Confirmed normal PRs already pass the executor-cloud preview build (only PRs that add a *new* DO migration hit 10211, which is rare and self-resolved now that the SQLite migration is applied). Not worth a permanent build-script workaround.

PR #1202, opened 2026-06-29
Fix floating auth tab strip for transparent-DCR OAuth connect modal

RhysSullivan:
## Problem

In the add-connection modal, the auth-method tab strip and its body live in a `Tabs` -> `TabsList` -> `TabsContent` group. The `TabsContent` card was wrapped in `{dcrActive ? null : (...)}`, so whenever an OAuth integration used transparent dynamic client registration (the MCP OAuth case) the card was suppressed.

The tab strip itself still renders whenever a custom-method `+` is available. The result: the OAuth tab plus the `+` floated with no card beneath it, and the tab strip's border looked detached (it jumped straight to "Connection saved to").

That same wrapper had a second symptom: the in-card "No app to choose" DCR explainer was written inside a `dcrActive ?` branch within the gated `TabsContent`, which the outer `dcrActive ? null` made permanently unreachable dead code.

## Fix

Always render the `TabsContent` card and let the inner `dcrActive` / `cimdActive` branches show their explainer. The tab strip gets its body back, and the DCR/CIMD explainer is now live instead of dead. The change is purely the removal of the `{dcrActive ? null : (...)}` gate; the rest of the diff is the block de-indenting by two spaces.

## Before / after

- Before: OAuth tab and `+` floating, no card, modal jumps to "Connection saved to".
- After: the OAuth tab sits on its "No app to choose / supports automatic setup" card.

## Verification

New selfhost browser scenario `e2e/selfhost/connection-modal-dcr-card.test.ts`: registers a remote MCP integration declaring an `oauth2` method (which makes it DCR-capable), opens the add-connection modal, and asserts the OAuth tab's body card renders under the tab strip. It passes on this branch and fails on the pre-fix code (the asserted copy was unreachable), so it guards the regression. Per-step screenshots and video are captured as run artifacts.

### DIFF
--- a/e2e/selfhost/connection-modal-dcr-card.test.ts
+++ b/e2e/selfhost/connection-modal-dcr-card.test.ts
@@ -1,0 +1,87 @@
+// Selfhost (browser, recorded): the add-connection modal for a transparent-DCR
+// OAuth MCP integration must render its body card, not just a floating tab
+// strip. A remote MCP integration that declares an oauth2 method is DCR-capable
+// (supportsDynamicRegistration), so the modal skips the app picker. When the
+// integration also offers the custom-method "+" the method tab strip renders.
+//
+// The regression this guards: the OAuth tab's TabsContent card (the "No app to
+// choose / automatic setup" explainer) was gated out whenever DCR was active,
+// leaving the tab strip with no card under it and a detached-looking border.
+// The same gate also made that explainer unreachable dead code. The per-step
+// screenshots are the artifact.
+import { randomBytes } from "node:crypto";
+
+import { Effect } from "effect";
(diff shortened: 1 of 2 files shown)


RhysSullivan:
### Before / after

Captured live from the selfhost app by the new e2e scenario (`e2e/selfhost/connection-modal-dcr-card.test.ts`), a remote MCP integration declaring an `oauth2` method (transparent DCR).

**Before** — the OAuth tab and `+` float with no card beneath; the modal jumps straight to "Connection saved to":

Before: floating OAuth tab strip with no body card

**After** — the OAuth tab sits on its "No app to choose / supports automatic setup" card:

After: OAuth tab anchored to its body card

PR #1204, opened 2026-06-29
Fix self-host/Cloudflare update card always showing; show version in the dashboard

RhysSullivan:
Fixes a user report: on a Docker (Railway) self-host, the sidebar "update available" card showed even when running the latest version. Same report asked to surface the version in the dashboard. Both come from one root cause.

## Root cause

The self-host and Cloudflare builds baked a **placeholder** `VITE_APP_VERSION` (`0.0.0-selfhost` / `0.0.0-cloudflare`) into the web shell. The update card compares that against the published `executor` dist-tags, so `compareVersions("0.0.0-selfhost", "1.5.23")` was always `-1` — permanently "behind", regardless of the deployed version. (The npm CLI and desktop builds already bake the real version, so they were fine.)

## Fix

- `apps/host-selfhost` and `apps/host-cloudflare` vite builds now read the real release version from the CLI package, exactly as `apps/local` does. The Docker image and Worker are built at the release tag, so this matches the published version.
- The multiplayer shell sidebar footer now shows `v<version>` (it showed no version before), so self-hosters can see what they are on. Hidden on builds that do not inject a version (e.g. cloud).

After the fix: a build that is current shows **no card** and displays its version; when a newer version is published, the card returns.

no card when current, version in footer

## Testing

- e2e `selfhost|cloudflare/update-card-render.test.ts` now has two scenarios each, both green on the real self-host and Cloudflare UIs:
  - behind -> the card shows (existing), and
  - current -> NO card, and the footer shows the running version (new regression test). With the old placeholder this test fails because the card would show even against an older published version.
- Format, lint, typecheck pass.

Note: deployed instances pick this up on their next image pull / redeploy (the version is baked at build time).

### DIFF
--- a/.changeset/selfhost-update-card-version.md
+++ b/.changeset/selfhost-update-card-version.md
@@ -1,0 +1,7 @@
+---
+"@executor-js/host-selfhost": patch
+"@executor-js/host-cloudflare": patch
+"@executor-js/react": patch
+---
+
+Fix the self-host and Cloudflare web dashboards showing "update available" even on the latest version. The builds baked a placeholder version (`0.0.0-selfhost` / `0.0.0-cloudflare`) into the shell, so the update check always compared as behind. They now bake the real release version, and the sidebar footer shows the running version so you can see what you are on.
--- a/apps/host-cloudflare/vite.config.ts
+++ b/apps/host-cloudflare/vite.config.ts
@@ -1,3 +1,4 @@
+import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 
 import { defineConfig } from "vite";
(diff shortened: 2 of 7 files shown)


RhysSullivan:
Verified against the **production build** (prod `serve.ts` over the `vite build` dist, the exact artifact the Docker image serves). The dist bakes the real version (`grep` finds `1.5.23`, no `0.0.0-selfhost` placeholder), and the e2e card-render scenarios pass on the `selfhost-docker` target attached to it: current build shows no card and the footer shows the version.

docker prod build: no card, v1.5.23 in footer

(A real container wraps this same build; the web bundle and version injection are identical.)

PR #1234, opened 2026-06-30
e2e CLI: keep the self-host data dir across restarts

RhysSullivan:
## Problem

`cli up selfhost` wiped the data dir on every boot. The dev CLI instance is meant to be persistent (it lives until `down`), so this erased a connected MCP client's dynamic OAuth registration on every restart, and the client then failed with `invalid_client` until it re-registered.

## Fix

Default `cli up` to keep the data dir; add `--fresh` to opt into a wipe. This matches the boot recipe's stated intent (the dev CLI is persistent). The vitest globalsetup is unaffected: it calls `bootSelfhost` directly with `fresh` defaulted on, so suite runs stay hermetic.


RhysSullivan:

PR #1239, opened 2026-07-01
Repair the self-host e2e suite and run it in CI

RhysSullivan:
## Why

The self-host e2e project (`vitest --project selfhost`) has never run in CI: the `test` job filters e2e out (`turbo run test --filter=!@executor-js/e2e`) and the only e2e job runs a single stdio scenario on push. With nothing exercising it, the suite drifted red as the app moved on. A full run showed 9 of 56 files failing, none of them a real regression in the tested behavior. This repairs them and adds a CI gate so it stays honest.

## Fixes

**Stale selectors (the connect modal now uses an affixed credential field).** The single-input bearer credential input renders an "Authorization: Bearer " affix with placeholder `token`, not the old `paste the value / token`. Updated the selectors, scoped to the dialog: `scenarios/connect-handoff.test.ts`, `scenarios/connect-handoff-session.test.ts`, `selfhost/auth-methods-ui.test.ts`.

**Racy action-bar position read.** `scenarios/openapi-add-integration-action-bar.test.ts` measured the Cancel button "in flight", but the floating action bar unmounts the instant the router navigates, so the read blocked the full timeout. The single-node counts already cover the reported doubled/ghosted-button regression; the step now asserts the submit commits and lands on the integration.

**Shared-admin connection count.** `scenarios/api-tools.test.ts` asserted a fresh identity has zero connections. That holds only on isolated-identity targets; self-host shares the bootstrap admin, so other scenarios' connections legitimately appear. The list call still runs everywhere (proving the endpoint); the zero-count assertion is gated off self-host per `e2e/AGENTS.md`.

**Multi-tenant-only 404.** `scenarios/org-slug-routing.test.ts` expected an unknown org slug to 404. That is a multi-tenant contract: self-host is single-tenant, `/account/me` always returns the instance org, and the slug is cosmetic. Gated the 404 step off self-host.

**Cloud-shaped toolkit URL.** `selfhost/toolkits-mcp.test.ts` connected to `/e2e-org/mcp/toolkits/...`. Self-host advertises the bare `/mcp/toolkits/...` (org prefixing is a cloud convention), and the server's RFC 9728 protected-resource doc reports the bare resource, which MCP SDK 1.29's stricter `selectResourceURL` requires the client URL to match. Connect to the URL self-host actually publishes.

**Quarantined (tracked follow-up).** `scenarios/microsoft-emulator.test.ts` and the Microsoft leg of `scenarios/oauth-client-handoff.test.ts` are skipped with a documented reason. `microsoft.addGraph` only accepts the canonical Microsoft Graph spec in the streamable block-YAML profile (it structurally splits the 37MB doc to avoid OOMing the 128MB Workers isolate) and hard-errors on anything else; the emulator serves a small spec outside that profile. Making these pass needs the emulator to serve a block-YAML Graph spec (or a non-Workers compile path), which is separate work. A `skip` option was added to the scenario helper so the quarantine stays visible in the report rather than silently deleted.

## CI

New `E2E (self-host)` job in `ci.yml`: boots the self-host target via its globalsetup (no external infra), installs Playwright Chromium, and runs `bun run test:selfhost` on pull requests and push. If the browser scenarios prove flaky on CI, gate it to push-only like `e2e-local`, or add a retry.

## New coverage

`selfhost/oauth-popup-callback-org-state.test.ts`: a browser-free regression test for the org-scoped OAuth popup callback (the bug fixed in #1235). It drives the org-context flow, asserts the provider state is wrapped, then does an authenticated GET of the callback and asserts it completes ("Connected") instead of "OAuth session expired or not found". No Playwright, so it is a good candidate to gate this class of regression cheaply.

## Verification

- Full self-host suite locally: 55 files pass, 1 skipped; 72 tests pass, 4 skipped, 0 fail.
- `oxlint --deny-warnings`, `oxfmt --check`, and `turbo run typecheck` (42/42) pass.

Note: the cross-target scenarios also run on the cloud target, which I could not boot locally. The org-slug and api-tools changes branch on the target so cloud behavior is unchanged, and the Microsoft quarantine applies on both targets.

## Relationship to #1235

The first commit cherry-picks #1235's OAuth popup fix (authorship preserved) so the OAuth-callback scenarios and the new test above are green and CI passes. If #1235 merges first, rebasing this branch drops the duplicate commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,44 @@ jobs:
         run: bunx vitest run --project local local/stdio-mcp.test.ts
         working-directory: e2e
 
+  e2e-selfhost:
+    name: E2E (self-host)
+    # Runs on PRs and push: the self-host project boots its own dev server (no
+    # external infra) and is the regression guard that PR CI was missing — the
+    # org-scoped OAuth callback bug lived exactly here and shipped green because
+    # nothing ran this suite. Browser scenarios are included; if they prove
+    # flaky on CI, gate this to push-only (like e2e-local) or add a retry.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.11
+
+      # The self-host web app + emulator OAuth flows spawn Node, and some
+      # scenarios drive a headless browser: pin Node 22 and install Chromium.
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      # Full fresh-checkout setup: installs deps AND builds the vite-plugin
+      # bundle + react console routes the web boot needs (a bare `bun install`
+      # leaves those unbuilt). bootstrap also fetches Chromium, but without the
+      # ubuntu system libs the headless shell needs — the step below adds
+      # `--with-deps` and the headless-shell download.
+      - run: bun run bootstrap
+
+      - name: Install Playwright Chromium (with system deps)
+        run: bunx playwright install --with-deps chromium chromium-headless-shell
+        working-directory: e2e
+
+      # Boots the self-host dev server via its globalsetup and runs the
+      # cross-target `scenarios/**` plus the selfhost-only `selfhost/**` suite.
+      - name: Run the self-host e2e suite
+        run: bun run test:selfhost
+        working-directory: e2e
+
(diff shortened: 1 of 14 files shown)


RhysSullivan:
Superseded on main: #1392 gates the self-host docker publish on the e2e suite and #1396 landed the selfhost-docker e2e fixes.

PR #1246, opened 2026-07-02
fix(mcp): keep tool catalogs in sync with the server's live tool set

RhysSullivan:
## Problem

An MCP connection's tool catalog was listed once at connection create and only re-listed on an executor-side config edit or a manual Refresh click. Server-side catalog changes (tools added, removed, renamed) were never picked up: stale tools kept failing invocations, and new tools never appeared.

Three spec gaps underneath:

- `tools/list` pagination was ignored — discovery called `listTools()` once and kept only the first page.
- `notifications/tools/list_changed` was not handled anywhere (no handler registered, no capability interest, and connections closed immediately after each call).
- Unknown-tool rejections on `tools/call` surfaced as generic errors with no self-healing.

There was also a latent data-loss bug: a transient outage during any re-list persisted an empty tool set, wiping working tools.

## Changes

**MCP plugin**
- Discovery follows `nextCursor` until exhausted (bounded at 100 pages against cycling cursors).
- Every invoke connection registers a `tools/list_changed` notification handler; a notification received during the call window marks the connection's persisted catalog stale, and the next tools read re-lists.
- Unknown-tool rejections are detected in both wire shapes — the spec's `-32602` protocol error and the reference TS SDK's `isError` envelope ("Tool X not found") — anchored to the exact tool name so domain errors can't false-positive. The call answers a typed `mcp_tool_unknown` failure telling the caller to re-list, and the catalog marks stale.
- Failed listings return `incomplete: true` instead of an empty catalog.

**Core SDK**
- New `ctx.connections.markToolsStale(ref)`: clears `tools_synced_at` so the next tools read re-produces the connection, without blocking the in-flight call.
- The stale-sync in `toolsList` now has three triggers: stale-marked (NULL stamp), config-revised (existing behavior), and TTL-expired — plugins that declare the new `remoteToolCatalog` flag (MCP does) are re-listed on read once older than `ExecutorConfig.toolsSyncTtlMs` (default 15 min, `null` disables). This covers servers that change tools without ever notifying. Steady state stays one indexed query returning nothing.
- `incomplete` results keep the previously persisted catalog rather than wiping it, and still stamp the sync time so a down server isn't re-dialed on every read.

## Scope note

Connections are dialed per-operation (Workers constraint), so `list_changed` notifications are only received during call windows; the TTL guarantees convergence for changes that land while no call is open. A persistent listener connection for long-running hosts would be a follow-up and can't work on Cloudflare, so this fix isn't coupled to it.

## Testing

New `catalog-sync.test.ts` covers all five behaviors end-to-end against a live MCP server whose catalog mutates at runtime (new `makeMutableCatalogMcpServer` fixture):

- `list_changed` during a call → stale mark → next read re-lists
- unknown-tool rejection → typed `mcp_tool_unknown` → catalog heals on next read
- TTL expiry re-lists on read
- fresh catalog inside the TTL serves persisted rows without dialing the server (asserted via session count)
- pagination across multiple `tools/list` pages

Full suite: 36/36 packages pass; format, lint, typecheck clean. Changeset included (`@executor-js/plugin-mcp` + `@executor-js/sdk` patch).

### DIFF
--- a/.changeset/mcp-tool-catalog-sync.md
+++ b/.changeset/mcp-tool-catalog-sync.md
@@ -1,0 +1,23 @@
+---
+"@executor-js/plugin-mcp": patch
+"@executor-js/sdk": patch
+---
+
+Keep MCP tool catalogs in sync with the server's live tool set. Previously a
+connection's tools were listed once at create time and never updated unless the
+integration's config changed or a user clicked Refresh, so server-side tool
+changes silently broke invocations.
+
+- `tools/list` discovery now follows `nextCursor` pagination per the MCP spec,
+  so servers with paginated catalogs list completely instead of first-page-only.
+- The client handles `notifications/tools/list_changed` received during a tool
+  call and marks the connection's persisted catalog stale; the next tools read
+  re-lists from the server.
+- An unknown-tool rejection from the server (protocol error or the reference
+  SDK's error envelope) returns a typed `mcp_tool_unknown` failure telling the
+  caller to re-list, and marks the catalog stale so it heals on the next read.
+- Remote catalogs now also refresh on read once older than a freshness TTL
+  (`ExecutorConfig.toolsSyncTtlMs`, default 15 minutes, `null` to disable),
+  covering servers that change tools without notifying.
+- A failed listing (server unreachable, auth not ready) no longer wipes the
+  previously persisted catalog; it is kept and retried after the TTL.
--- a/e2e/scenarios/mcp-catalog-sync-ui.test.ts
+++ b/e2e/scenarios/mcp-catalog-sync-ui.test.ts
@@ -1,0 +1,111 @@
+// Cross-target (browser): MCP tool-catalog freshness, FILMED through the real
+// web UI. The session video + per-step screenshots are the artifact: a user
+// adds a live MCP server, sees its tool in the Tools tab, runs a tool that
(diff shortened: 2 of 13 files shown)


RhysSullivan:
Added end-to-end proof: [`e2e/scenarios/mcp-catalog-sync.test.ts`](https://github.com/RhysSullivan/executor/blob/mcp-tool-catalog-sync/e2e/scenarios/mcp-catalog-sync.test.ts) — four cross-target scenarios (cloud + selfhost, 4/4 passing on both) driving only public surfaces against real MCP servers whose catalogs mutate mid-scenario:

1. **list_changed** — a tool call renames the server's catalog mid-call; the notification rides the open connection and the very next `tools.list` read serves the renamed catalog. No refresh, no failing retry.
2. **unknown-tool self-heal** — the catalog mutates outside any call window; invoking the retired tool fails with the typed `mcp_tool_unknown` error and that failure alone heals the catalog for the next read.
3. **pagination** — a server paging `tools/list` via `nextCursor` registers its whole catalog (a first-page-only client silently loses the second page).
4. **outage resilience** — a refresh against a dead (503) server keeps the working catalog instead of wiping it, then converges to the server's changed catalog once it recovers.

Scenarios 3 and 4 use a purpose-built minimal streamable-http JSON-RPC fixture (mutable pages + kill switch); 1 and 2 run against the real SDK `McpServer`.

PR #1249, opened 2026-07-02
Carry the verified Principal into DO agent sessions

capeflow:
## Problem

The two session stores give the host's server builder different views of the authenticated caller.

The in-memory store hands `buildServer` the full `Principal`:

```ts
// packages/hosts/mcp/src/in-memory-session-store.ts
return buildServer(principal, { ...buildOptionsFor(request, () => createdSessionId), resource })
```

The Durable Object bridge does not. `propsForPrincipal` in the cloud and host-cloudflare agent handlers builds `McpSessionInit` from just `principal.organizationId` and `principal.accountId`, and `SessionMeta` can never contain more than `McpSessionInit` carried in. Everything else the auth seam verified (email, name, avatarUrl, roles, organizationSlug) is discarded at that boundary, even though the worker had it in hand one call earlier.

So a host's `buildMcpServer` structurally cannot see the identity the request was authenticated with. A server builder that reads the principal (per-caller tool policies, audit attribution) behaves correctly on the in-memory store and silently loses identity on DO sessions.

## Change

Carry the principal whole across the DO boundary, the same way `webOrigin` is already carried:

- `McpSessionInit` and `SessionMeta` gain one optional `principal?: Principal` field. No new type: `Principal` is already the shared authenticated-caller noun on both sides of the seam.
- The base copies it onto the stored meta alongside `webOrigin`. That merge is extracted into a small pure `carrySessionInit` helper (`session-meta.ts`) so it is testable and so no host's `resolveSessionMeta` needs to know about either carried field.
- The cloud and host-cloudflare agent handlers add `principal` to the session props they already build from that same principal.

Optional and additive: unset changes no behavior, and ownership validation stays keyed on userId + organizationId only.

Two semantics made explicit on the field docs:

- **Snapshot at mint.** `SessionMeta` persists in DO storage, so the carried principal is frozen at session create, exactly like `userId` and `organizationId` today (a session does not re-authenticate). That matches the in-memory store, which also captures the principal at create.
- **PII at rest.** Email/name/avatarUrl land in DO storage for the session's lifetime (bounded by the session alarm). The in-memory store already holds the same data for the same lifetime, in process memory. The optional shape keeps this opt-in per host if that difference matters.

Context: we run the Cloudflare host with host-side plugins built per session in `buildMcpServer`; one enforces access grants matched on the caller's email and groups, which works against the in-memory store's contract and needed this carrier on the DO path.

## Testing

All run on this branch, all passing:

- `bun run --cwd packages/hosts/cloudflare test -- src/mcp/session-meta.test.ts`: new `carrySessionInit` test, 4/4 (carry-through, both fields, unset leaves no phantom keys, no clobbering of host-resolved meta)
- `bun run --cwd apps/host-cloudflare test -- src/worker.e2e.node.test.ts`: workerd/miniflare e2e through the real DO session path (initialize, cross-request survival, tools/call), 8/8
- `bun run --cwd apps/cloud test -- src/mcp-session.e2e.node.test.ts`: cloud session harness, 4/4
- Full suites of the touched packages: hosts/cloudflare 15/15, hosts/mcp 57/57, cloud 191/191
- `bunx turbo run typecheck` on the three touched packages (24/24), `bun run format:check` and `bun run lint` clean at the root

### DIFF
--- a/apps/cloud/src/mcp/agent-handler.ts
+++ b/apps/cloud/src/mcp/agent-handler.ts
@@ -105,6 +105,9 @@ const propsForPrincipal = (
         elicitationMode: readElicitationMode(request),
         resource,
         webOrigin: new URL(request.url).origin,
+        // The whole verified caller, so the session DO rebuilds its runtime
+        // with the same identity the in-memory store passes to buildServer.
+        principal,
       },
       propagation,
     };
--- a/apps/host-cloudflare/src/mcp/agent-handler.ts
+++ b/apps/host-cloudflare/src/mcp/agent-handler.ts
@@ -96,6 +96,9 @@ const propsForPrincipal = (
         // resource.
         resource: defaultMcpResource,
         webOrigin: new URL(request.url).origin,
+        // The whole verified caller, so the session DO rebuilds its runtime
+        // with the same identity the in-memory store passes to buildServer.
+        principal,
       },
       propagation,
     };
--- a/packages/hosts/cloudflare/src/mcp/agent-session-durable-object.ts
+++ b/packages/hosts/cloudflare/src/mcp/agent-session-durable-object.ts
@@ -14,7 +14,7 @@ import {
   PAUSED_APPROVAL_TIMEOUT_MS,
   type PausedExecutionHooks,
 } from "@executor-js/host-mcp/tool-server";
-import { defaultMcpResource, type McpResource } from "@executor-js/host-mcp";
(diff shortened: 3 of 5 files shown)


capeflow:
CleanShot 2026-07-02 at 08 53 49@2x


CleanShot 2026-07-02 at 08 54 12@2x

CleanShot 2026-07-02 at 09 27 40@2x

CleanShot 2026-07-02 at 09 32 00@2x

PR #1250, opened 2026-07-02
Source integration favicons from integrations.sh/logo

RhysSullivan:
integrations.sh now exposes `GET /logo/{domain}?sz=N`, an edge-cached proxy over a proper logo API, so favicons come from there instead of hitting Google's favicon service directly.

- `integrationFaviconUrl` now builds `https://integrations.sh/logo/{domain}?sz=N`.
- The resolution cascade gains a final step, `integrationFaviconFallbackUrl` (the old Google `s2/favicons` URL), so if the proxy is unreachable the `` onError handler walks to it automatically; the BoxIcon placeholder remains the terminal fallback.
- Domain extraction and local-URL guards are unchanged (still tldts registrable-domain based).

Tests updated and passing (14), full `bun run typecheck` green.


RhysSullivan:
Extended the sweep: preset icons that pointed at vendor favicon.ico paths (stripe.com, linear.app, notion.com, ... 28 URLs across the openapi/graphql/mcp/microsoft plugins) now route through integrations.sh/logo too. Kept as-is: svgl.app and fonts.gstatic product logos (per-product marks for Google/Microsoft sub-products that a domain-keyed proxy can't distinguish), bundled marketing assets, and the app's own favicon links. Verified every swapped domain serves a real image via the proxy; plugin sdk tests and typecheck pass.

PR #1256, opened 2026-07-02
executor generate: export the tool catalog as OpenAPI (or a typed TS client), backed by REST tool invocation

RhysSullivan:
## What

`executor generate` turns an Executor instance into a typed API proxy. The primary artifact is an **OpenAPI 3.1 document** describing every visible tool as one REST operation, so you plug your catalog into whatever client generator you already use:

```bash
executor generate                        # writes executor.openapi.json
npx openapi-typescript executor.openapi.json -o executor-api.ts
```

The same document is served live at `GET /api/tools/export/openapi` (bearer auth), and its operations are real endpoints: `POST /api/tools/invoke/{tool path}` runs the tool through Executor's execution engine, so credentials, policies, and approvals all apply. Responses are a uniform envelope: `{ ok: true, data }` or `{ ok: false, error }`; approval-gated calls return `error.code: "execution_paused"` with an `executionId` and `resumePath`; unknown tools answer 404.

Prefer a ready-made client? `--format typescript` emits a single self-contained TypeScript file (dependency-free Proxy runtime + full per-tool input/output types), and `--format both` writes both artifacts. The TS client speaks the same invoke endpoint the OpenAPI document describes.

```ts
import { createExecutorClient } from "./executor.gen";
const executor = createExecutorClient();
const created = await executor.github.org.main.issues.create({ title: "Hi" });
if (created.ok) console.log(created.data.number); // fully typed
```

## How

- **`tools.export` + `GET /tools/export`**: bulk schema-bearing catalog read (input/output JSON schemas plus each connection's shared `$defs` trimmed to the referenced subset, policy-filtered). One request regardless of catalog size.
- **`generateOpenApiSpec`** (sdk): catalog → OpenAPI 3.1. One POST operation per tool at `/tools/invoke/{path}`, request body from the tool's input schema, envelope responses, shared `$defs` hoisted once into `components.schemas` under per-connection namespaces (a 10k catalog references each shared schema once instead of inlining 10k copies), unique identifier-safe operationIds, per-connection tags.
- **`POST /tools/invoke/:path`**: direct REST invocation through the execution engine (same pause/approval semantics as every other host surface), with a sandbox sentinel so a missing tool surfaces as HTTP 404 instead of a buried execution error string.
- **`compileToolChunkTypeScript`** (sdk): compiles ~200 tools per json-schema-to-typescript pass for the TS artifact. Whole-catalog single passes are super-linear (30s+ at 10k tools); chunking does the same work in ~400ms, and a failing chunk retries tool-by-tool so one broken schema degrades to `unknown`.
- **CLI**: `executor generate --format openapi|typescript|both` (default openapi) with `--output`, `--integration`, `--connection`, `--include-static`, on the existing server-target/auth machinery.

## Scale (per the requirement: 10,000+ tools toward the total, multiple specs, emulate)

The scale test builds the catalog the way real instances do: **many specs**, not one giant one. It registers real service specs served by `@executor-js/emulate` emulators (github, stripe) by URL — the exact path a user adding those integrations takes — plus a synthetic fleet of 8 specs topping the combined catalog past 10,000 tools. Then, over the combined catalog:

- `tools.export` returns everything in one read (~150ms),
- `generateOpenApiSpec` emits the document with one path per tool,
- the **real `openapi-typescript` generator** consumes that document and its output contains all 10k+ invoke paths — third-party interop proven, not assumed,
- `generateToolProxySource` emits the TS client and the whole file typechecks under `--strict`, including a consumer snippet against a mid-fleet tool.

Time tripwires on export/spec-gen/ts-gen catch any regression to per-tool or whole-catalog compilation.

## Testing

- `packages/core/sdk/src/specgen.test.ts`: operation shapes, namespaced component refs, cross-connection name isolation, operationId dedupe.
- `packages/core/sdk/src/typegen.test.ts`: generated TS typechecked with the real compiler (valid calls accept, wrong types reject); the transpiled runtime run against a fake fetch (ok unwrap, paused error with approval URL, path-injection rejected before any request); naming edge cases; `tools.export` grouping/trimming/policy filtering.
- `apps/local/src/tools-invoke.test.ts`: invoke + export endpoints over real HTTP handlers end to end — spec registered, connection created, a real upstream echo server dialed with the rendered credential; pause-with-resume-coordinates; 404; non-object body rejection; served document's `servers[0]` derivation.
- `packages/plugins/openapi/src/sdk/typegen-scale.test.ts`: the multi-spec 10k scale test above.
- Verified live against a running daemon: generated the OpenAPI document via CLI, ran `openapi-typescript` on it, and made a spec-driven call (base URL and path taken from the document itself) that returned real data over `POST /api/tools/invoke/...`.

Full gates run: format:check, lint, typecheck, test. Pre-existing flake note: `host-selfhost`'s `scope-isolation.test.ts` timeout fails identically on a clean checkout without this branch.


aryasaatvik:
looking forward to this being able extract large amounts of data by composing integrations outside the model context would be massive

PR #1271, opened 2026-07-02
fix: add google discovery media upload operations

batumilove:
## Summary

Fix Google Discovery media upload modeling so upload-capable methods expose a binary request body instead of metadata-only JSON.

- Decode `supportsMediaUpload` and `mediaUpload` from Google Discovery method definitions.
- Generate separate simple media upload operations, for example `files.createMedia` and `files.updateMedia`.
- Use the Discovery simple upload path under `/upload/...`.
- Preserve the method HTTP verb, so `files.createMedia` is `POST` and `files.updateMedia` is `PATCH`.
- Add `uploadType=media` and an `application/octet-stream` request body, which is exposed as `bodyBase64` by existing OpenAPI extraction.
- Preserve the original metadata JSON operation unchanged.
- Cover both single Discovery conversion and bundled Google Discovery conversion.

## Why

Drive `files.create` advertised media upload support, but Executor generated only the metadata JSON operation. Calling it with `uploadType=media` had no binary/media body path and produced 0-byte Drive files.

## Validation

- `bun run --cwd packages/plugins/google test -- src/sdk/discovery.test.ts`
- `bun run --cwd packages/plugins/google test`
- `bun run --cwd packages/plugins/openapi test -- src/sdk/non-json-body.test.ts src/sdk/index.test.ts`
- `bun run format`
- `bun run format:check`
- `bun run lint`
- `bunx tsgo --noEmit`
- `git diff --check`

### DIFF
--- a/packages/plugins/google/src/sdk/discovery.test.ts
+++ b/packages/plugins/google/src/sdk/discovery.test.ts
@@ -395,6 +395,195 @@ it.effect("supplies documented scopes when Picker Discovery omits auth metadata"
   }),
 );
 
+it.effect(
+  "generates a separate media-upload operation for Google Discovery methods with supportsMediaUpload",
+  () =>
+    Effect.gen(function* () {
+      const result = yield* convertGoogleDiscoveryToOpenApi({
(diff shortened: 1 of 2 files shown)


RhysSullivan:
thanks!

PR #1283, opened 2026-07-03
Show the daemon as "Executor" in Login Items instead of the developer name

RhysSullivan:
## Problem

macOS 13+ groups bare legacy LaunchAgents in System Settings → Login Items & Extensions under the **signing certificate developer name**. Executor installs its daemon as `~/Library/LaunchAgents/sh.executor.daemon.plist` with no app association, so every user sees a "Rhys Sullivan" login item (with a generic icon) instead of "Executor".

## Fix

`generateLaunchdPlist` now emits Apple's supported association key when (and only when) the service binary lives inside a macOS app bundle:

```xml
<key>AssociatedBundleIdentifiers</key>
<array>
  <string>sh.executor.desktop</string>
</array>
```

The bundle id is read from the enclosing bundle's `Contents/Info.plist` (with a constant fallback for `Executor.app` if the read fails). Standalone CLI installs (npm/homebrew, binary not inside a `.app`) produce a byte-identical plist to before. The app and daemon already share a Team ID (the daemon ships inside the bundle), which is what BTM requires to honor the association.

## Verification

| Before — plist without the key | After — plist with the key |
|---|---|
| Login Items grouping the daemon under "Rhys Sullivan" with a generic icon | Login Items showing "Executor" with the app icon |
| Grouped under the signing cert developer name — "Rhys Sullivan, 1 item", generic grid icon. | Same daemon attributed to **Executor** with the real app icon. |

Reproduced and proved in a disposable tart macOS VM (Sequoia) using the real Developer-ID-signed release bundle:

- **Before** (`sfltool dumpbtm`): daemon record `Type: legacy agent`, `Parent Identifier: Rhys Sullivan`; Login Items shows a "Rhys Sullivan / 1 item" entry with a generic icon.
- **After** adding the key and re-registering: daemon record gains `Assoc. Bundle IDs: [ sh.executor.desktop ]`, `Parent Identifier: 2.sh.executor.desktop`; Login Items shows **Executor** with the app icon.
- Existing installs self-heal on the first fresh BTM ingestion of the updated plist (the app reinstalls the service on upgrade, rewriting the plist); no `sfltool resetbtm` needed for new installs since first ingestion already carries the key.
- Unit tests cover both branches (association emitted from a fixture bundle's Info.plist; key absent for standalone paths) plus the pre-existing plist invariants. `bun run --cwd apps/cli test` 18/18, typecheck clean.

### DIFF
--- a/apps/cli/src/service.test.ts
+++ b/apps/cli/src/service.test.ts
@@ -1,4 +1,7 @@
-import { describe, expect, it } from "@effect/vitest";
+import { afterEach, describe, expect, it } from "@effect/vitest";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 
 import {
   cmdSetValue,
@@ -11,6 +14,14 @@ import {
   parseNetstatListenerPids,
   parseSchtasksRunning,
 } from "./service";
+
+const tempDirs: Array<string> = [];
+
+afterEach(() => {
+  for (const dir of tempDirs.splice(0)) {
+    rmSync(dir, { recursive: true, force: true });
+  }
+});
 
 describe("service unit generation", () => {
   const launchdInput = {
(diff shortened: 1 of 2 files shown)


RhysSullivan:
## Before / after proof (tart macOS VM, real signed bundle)

Reproduced in a disposable `tart` Sequoia guest using the unmodified Developer-ID-signed `Executor.app` and the production `service install` flow. The only difference between the two states is the `AssociatedBundleIdentifiers` key this PR makes the CLI emit.

| Before — plist without the key | After — plist with the key |
|---|---|
| Login Items grouping the daemon under "Rhys Sullivan" with a generic icon | Login Items showing "Executor" with the app icon |
| Grouped under the signing cert's developer name — "Rhys Sullivan, 1 item", generic grid icon; the "Background Items Added" toast also names the developer. | Same daemon now attributed to **Executor** with the real app icon. (Toast text in the corner is a stale notification from the earlier registration.) |

Matching `sfltool dumpbtm` records from inside the VM:

**Before**
```
Identifier:        8.sh.executor.daemon
Executable Path:   /Applications/Executor.app/Contents/Resources/executor/executor
Type:              legacy agent (0x10008)
Parent Identifier: Rhys Sullivan
```

**After**
```
Identifier:        8.sh.executor.daemon
Executable Path:   /Applications/Executor.app/Contents/Resources/executor/executor
Assoc. Bundle IDs: [ sh.executor.desktop ]
Parent Identifier: 2.sh.executor.desktop
```

Notes:
- Signature verified in-VM before testing: `Developer ID Application: Rhys Sullivan (A7SQ9BT93Y)`, notarization accepted.
- The VM needed a BTM reset + reboot only because the item was **already registered** when the key was added in place. New installs ingest the plist with the key on first registration, and existing users self-heal when the app's next upgrade reinstalls the service and rewrites the plist.
- Screenshots live on the orphan branch `assets/login-items-proof`, not in this PR's diff.

PR #1300, opened 2026-07-04
Make paused-execution resume survive MCP session changes

RhysSullivan:
## Problem

When a paused execution's `resume` call arrives on a different MCP session than the one that created the pause (clients like Cursor recycle their MCP session between requests), the resume lands on a fresh session DO whose engine has no knowledge of the executionId. The client gets a generic "execution not found / may have expired" error — even seconds after the pause — and the original pause later auto-declines via the 4-minute pending-approval lease. Agents mis-diagnose this as a too-short expiry window and guess-and-retry.

Confirmed in production traces: pause at 17:07:09, new `McpSessionDO.init` at 17:07:18, resume miss 9 seconds after the pause, orphaned `Tool approval declined` (duration exactly 4m0s) at 17:11.

## Fix

Make model-mode `resume` session-portable:

- **Execution-owner directory**: a small DO (`McpExecutionOwnerDirectoryDO`, addressed `idFromName(executionId)`) records `{ owner session route, accountId, organizationId, expiresAt }` when a pause starts its pending-approval lease. Entries expire via alarm; expired reads are treated as absent. All directory reads/writes degrade gracefully — an unavailable directory never fails an execute/pause/resume.
- **Cross-session forward**: on a local resume miss, the tool-server's new optional `resumeFallback` hook looks up the owner and calls a new non-forwarding `resumeExecutionForModel` RPC on the owning session DO. That RPC re-validates the caller identity (account + organization from session meta, never client-supplied), runs the normal resume lifecycle — clearing the 4-minute lease so no later `approval_timeout` decline fires — and returns the continued execution result synchronously. A 10s deadline on the forward maps a wedged owner to `execution_expired`. Recursion-safe: an owner route pointing at the current session short-circuits.
- **Honest errors**: the generic miss result is replaced by distinct statuses — `execution_not_found`, `execution_expired`, `execution_forbidden`, `execution_already_settled` — with text that states the approval window (derived from the lease constant, not hardcoded).
- **Plannable deadlines**: paused payloads now carry `expiresAt`/`ttlMs` (threaded from the host lease duration) and mention the deadline in the pause text, so agents can schedule the resume instead of guessing.

Hosts that don't wire the fallback (local/selfhost stdio) keep exactly today's behavior. No timer values changed.

## Tests

- Two-session routing exercised through the real `resumeFromExecutionOwnerDirectory` → `resumeExecutionForModel` paths (forwarding stubbed only at the transport boundary): cross-session resume returns the owner's result, the owner's lease is released with no later auto-decline, identity mismatch is rejected without touching the owner engine, no recursion.
- Shared `mcpSessionDurableObjectName` helper test covers the DO-name derivation used by both apps (mutation-checked: breaking the derivation fails the suite).
- Directory DO unit tests (put/get/expiry/alarm), distinct-error-status coverage, same-session resume regression coverage.
- `bun run typecheck` 42/42; hosts/mcp 64 tests; hosts/cloudflare 24 tests.

## Deploy notes

New DO class `McpExecutionOwnerDirectoryDO` with additive `new_sqlite_classes` migrations in both apps (cloud `v4`, host-cloudflare `v2`); binding `MCP_EXECUTION_OWNER` is optional in code, so deploy order is safe for existing sessions.

### DIFF
--- a/apps/cloud/src/env-augment.d.ts
+++ b/apps/cloud/src/env-augment.d.ts
@@ -38,6 +38,7 @@ declare global {
       // workers and older local setups run without the binding, and the
       // limiter degrades to disabled when absent.
       EXECUTION_RATE_LIMITER?: import("@cloudflare/workers-types").DurableObjectNamespace;
+      MCP_EXECUTION_OWNER?: import("@cloudflare/workers-types").DurableObjectNamespace;
 
       // Optional per-org hourly execution rate-limit override, parsed as an
       // integer (defaults to EXECUTIONS_PER_ORG_PER_HOUR = 1000 when unset or
--- a/apps/cloud/src/mcp/agent-handler.ts
+++ b/apps/cloud/src/mcp/agent-handler.ts
@@ -13,6 +13,7 @@ import {
   withVerifiedIdentityHeaders,
 } from "@executor-js/cloudflare/mcp/do-headers";
 import type { McpSessionProps } from "@executor-js/cloudflare/mcp/agent-durable-object";
+import { mcpSessionDurableObjectName } from "@executor-js/cloudflare/mcp/execution-owner-directory";
 
 import { wrapMcpSseResponse } from "../observability/memory-metrics";
 import { cloudMcpAuth } from "./auth-provider";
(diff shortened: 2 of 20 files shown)


RhysSullivan:
Note on the red `Workers Builds: executor-cloud` check: that build is a *versioned* (preview) upload, and Cloudflare rejects version uploads containing Durable Object migrations (error 10211 — migrations require a non-versioned deployment). Any PR adding a DO class fails this check by construction; #1274 (ExecutionRateLimiterDO) failed it identically and deployed fine from main. The build itself compiles — the log shows the new `MCP_EXECUTION_OWNER` binding resolving — it only dies at the versions-upload step. The merge-to-main deploy applies the migration for real.

PR #1304, opened 2026-07-04
[codex] Wire Azure Trusted Signing for desktop releases

RhysSullivan:
## What changed

- Adds conditional `win.azureSignOptions` to the desktop Electron builder config.
- Wires Azure Trusted Signing environment variables into the Windows desktop publish leg.
- Adds a Windows-only fail-fast check so release builds do not silently publish unsigned Windows installers when signing config is incomplete.
- Keeps e2e Windows packages explicitly unsigned even if a developer shell has Azure signing env vars set.

## Validation

- `bun run --filter @executor-js/desktop typecheck`
- `bun run typecheck`
- Imported the builder config with and without signing env vars to verify conditional behavior.
- Parsed `.github/workflows/publish-desktop.yml` as YAML.


RhysSullivan:

PR #1311, opened 2026-07-04
Group connections into provider accounts with login_hint

RhysSullivan:
Adds a derived provider-account grouping (no new storage): connections on Google/Microsoft integrations group by probed account email, rendered as one account with its per-service connections. From an account row, additional services can be enabled one consent at a time; each consent carries login_hint so Google/Microsoft pre-select the account. Connection lists now expose a narrow lastHealth projection (status/identity/checkedAt) for the accounts view.

Independently reviewed (ship-with-nits; all three nits addressed: narrowed list projection, per-account connection naming, no-auto-chaining handler tests).


RhysSullivan:
Closing: superseded by the openapi catalog redesign. Connection identity now comes from the OAuth grant (id_token) rather than probed health-check identity, so the account grouping would need rebuilding on that foundation if we want it later.

PR #1321, opened 2026-07-04
fix: scope connection health refreshes

zrm625:
## Summary

- keep manual health-check results scoped to the clicked connection row
- refresh connection reads after the row applies its returned verdict
- update the comments around broad vs targeted connection refreshes

## Context
found in `v1.5.28`
The repro used a profile check on the test connection. Checking one row could briefly make every connection look like the same profile/account, even though the saved verdict belonged only to the clicked connection when using `GET gmail.users.getProfile`

--
Tried to document the footgun, but it may have lead to more change in files that were not in the direct change

### DIFF
--- a/e2e/scenarios/health-check-scoped-refresh.test.ts
+++ b/e2e/scenarios/health-check-scoped-refresh.test.ts
@@ -1,0 +1,397 @@
+// Cross-target (browser): manual connection health checks must update only the
+// row that was checked. A broad refresh used to refetch sibling account rows,
+// letting unrelated persisted identity changes appear as collateral.
+import { randomBytes } from "node:crypto";
+import { createServer } from "node:http";
+
+import { expect } from "@effect/vitest";
+import { Effect } from "effect";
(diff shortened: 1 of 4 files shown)


RhysSullivan:
claude - recordings

E2E repro (two connections on one integration: an org-owned "Workspace" row for alice, a user-owned "Personal" row for bob, distinct identities from a stub API).

**Before (main):** Check now on the Workspace row broadly invalidates the connections cache, refetching the Personal row too — it visibly picks up an unrelated identity change:

before

**After (this PR):** same flow, only the clicked row's read surfaces refresh; the Personal row keeps its cached identity:

after

The test asserts the visible bleed (the sibling row's identity text must never change during a check of the other row) and the network behavior (no `/api/connections?owner=user` refetch when checking an org-owned connection). Fails on main with:

```
AssertionError: Checking Alice refetched Bob's row and showed its offscreen identity.
```

and passes on this branch.

PR #1323, opened 2026-07-04
Support env-backed CLI server headers

### LINKED ISSUE #1322: Allow CLI server profiles to send env-backed HTTP headers for Cloudflare Access
Cloudflare is one of Executor's documented self-host/deployment paths, and the Cloudflare host docs already cover Access-gated `/mcp` usage with service-token headers for headless clients.

That works for direct MCP clients, but the `executor` CLI cannot currently use the same Access-protected instance because commands like:

```bash
executor login --base-url https://executor.example.com
```

must first call `/api/auth/cli-login`, and there does not appear to be a way for the CLI/server profile to attach required Cloudflare Access headers:

- `CF-Access-Client-Id`
- `CF-Access-Client-Secret`

This means a Cloudflare Access-protected self-hosted Executor can be used by MCP clients that support custom headers, but not cleanly by the Executor CLI itself.

Requested shape:

- Let `executor server add` / server profiles define HTTP headers.
- Prefer env-backed values, not plaintext persistence, for example:
  - `CF-Access-Client-Id=$EXECUTOR_CF_ACCESS_CLIENT_ID`
  - `CF-Access-Client-Secret=$EXECUTOR_CF_ACCESS_CLIENT_SECRET`
- Apply those headers consistently to CLI discovery/login and later API calls:
  - `/api/auth/cli-login`
  - `whoami`
  - `call`
  - `generate`
  - other remote server commands

This would make the blessed Cloudflare path usable end-to-end from both MCP clients and the official CLI, without requiring users to weaken Access or run a local header-injecting proxy.

🤖 *This content was generated with AI assistance using gpt-5.5.*

ramarivera:
## Summary

- Add env-backed HTTP headers to Executor server connection profiles.
- Add `executor server add --header-env Header-Name=ENV_VAR` for Cloudflare Access service token style deployments.
- Apply resolved profile headers to remote API calls, CLI login discovery, device code requests, token polling, and token refresh.
- Preserve header mappings across login, logout, and OAuth refresh.
- Scope device and token endpoint headers to the Executor server origin so profile headers are not sent to third party OAuth providers.

Fixes #1322

## Testing

- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run --cwd packages/core/sdk test src/server-connection.test.ts`
- `bun run --cwd apps/cli test src/server-profile.test.ts src/device-login.test.ts`
- `bun run --cwd e2e vitest run --project local local/server-profile-header-env.test.ts`

`bun run test` was also attempted. It failed in the existing `@executor-js/host-mcp` stdio integration because this machine already has local processes listening on `localhost:4788`; the test connected to that ambient daemon and got `Unauthorized`. The changed SDK and CLI tests passed, and full typecheck passed.

## Live smoke

Against `https://executor.ai.roxasroot.net/api/auth/cli-login`:

- Without Cloudflare Access service token headers: response was Cloudflare Access sign-in HTML.
- With service token headers resolved from env: request passed Access and reached the app, which returned 404 because the currently deployed instance does not expose `/api/auth/cli-login`.


RhysSullivan:
Superseded by #1539, which replayed this branch's commit (preserving authorship) onto current main and has now merged. Thanks @ramarivera for the original implementation.

PR #1324, opened 2026-07-04
Reconnect updates the existing DCR connection

RhysSullivan:
Reconnecting an expired DCR OAuth connection created a second connection instead of re-authing the existing one: the flow went through the create path with the old name as a display label, which got re-normalized into a new slug. Reconnect now carries an exact connection reference through the OAuth session so the same row is updated in place, with a typed error if the connection was removed mid-flow.

Includes an e2e repro for the duplicate, plus an expected-failure repro for a related symptom: after a successful reconnect the row keeps showing Expired until a hard refresh (the mounted health probe isn't cleared). That fix will follow.


RhysSullivan:
Superseded on main: #1316 fixed the DCR reconnect handoff, with further DCR registration rework in #1443/#1456.

PR #1326, opened 2026-07-05
Unwrap Effect causes before Sentry capture

RhysSullivan:
Frontend error reports were capturing raw Effect Cause objects, so Sentry showed a generic placeholder instead of the underlying error and grouped everything together. Unwrap causes into real errors at the capture boundary and attach the pretty-printed cause as context.


RhysSullivan:

PR #1327, opened 2026-07-05
Fix Windows updater signing config

RhysSullivan:
The publish workflow passed the mac signing env to every build leg, so Windows releases were signed with the wrong certificate and clients reject updates. Scope the mac cert to mac legs, add dedicated Windows signing env, fail CI packaging when the Windows cert is missing, and handle updater download rejections instead of letting them surface as unhandled errors.


RhysSullivan:

PR #1332, opened 2026-07-06
Custom tools: define tools in a GitHub repo and sync them into the catalog

RhysSullivan:
## What

Users can point Executor at a GitHub repo containing `tools/*.ts` files (`defineTool` default exports) and sync them into the org tool catalog. Synced tools run in a QuickJS sandbox, declare the integrations they need, and invoke through the normal policy/credential path — the caller picks which connection serves each declared integration at invoke time.

Console: a Custom tools tile in the connect-integration modal (repo + ref + GitHub connection), with a source panel showing published tools, skipped files, and a Sync button. Sync is manual; the repo is the source of truth and git history is the version history.

## How

- `packages/plugins/apps`: publish pipeline (discover → bundle → collect → commit; descriptor written into the snapshot, commit last, CAS ref updates), git-backed artifact store, per-scope libSQL db, QuickJS sandbox behind a serializable bridge, tenant-partitioned storage. `input`/`output` accept Standard Schema (JSON-schema export via `~standard.jsonSchema`, validation via `~standard.validate` in-sandbox) or a raw JSON schema object; zod is bundled. One tool per file; identity is the file path.
- GitHub ingest via the REST tree API behind a DI seam (payload limits enforced before fetching blob bodies; symlinks/submodules skipped; re-sync of an unchanged SHA is a no-op).
- Self-host wiring: apps plugin registered in the host plugin list, per-request resolver routing bridged calls through `ctx.execute` (policy, credentials, approvals — the caller's elicitation context is threaded through), authenticated sync route.
- `workflows/`, `ui/`, and `skills/` folders in a repo are accepted and reported as not supported yet, so repos written now keep working as those land.

## Testing

- Package suites: 54 tests (pipeline atomicity/determinism, CAS races, strict bridge dispatch, binding errors, tenant isolation, symlink filtering — each review finding has a fail-before/pass-after regression test).
- host-selfhost: 81 tests including a wire e2e that boots the real app composition and, over real HTTP against the GitHub emulator: 401 without auth, sync → publish, catalog listing with the projected connection enum, invoke through real QuickJS with the emulator ledger proving the upstream call, idempotent re-sync, add/remove churn.
- Walked live against a real private GitHub repo end to end (sync, invoke with real data, add/remove a tool upstream).


RhysSullivan:
Superseded by the apps stack: #1361, #1362, #1363, #1365.

PR #1359, opened 2026-07-08
Separate catalog-sync status from connection health

RhysSullivan:
A failed tool sync used to persist a degraded health verdict on the connection, so a single transient MCP dial failure or spec-blob read hiccup showed DEGRADED across whole connection lists even though the credentials were fine.

Sync trouble now lives on its own connection column (`tools_sync_error`) with a consecutive-failure count and the plugin's reason; `last_health` is written by health probes only. The UI debounces the sync signal (3+ consecutive failures) and shows it as a muted stale-catalog hint instead of the health treatment. Existing tool-sync pseudo-verdicts are cleared by migration (cloud drizzle 0010 + a stamped sqlite data migration), and fresh `unknown` verdicts no longer re-probe on every page load.

### DIFF
--- a/.changeset/health-sync-separation.md
+++ b/.changeset/health-sync-separation.md
@@ -1,0 +1,5 @@
+---
+"executor": patch
+---
+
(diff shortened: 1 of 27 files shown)


RhysSullivan:
Closing in favor of #1612

PR #1360, opened 2026-07-08
Surface GraphQL introspection errors instead of silent empty tool sets

RhysSullivan:
Adding a GraphQL source whose endpoint responds with something other than a valid introspection result (wrong URL, auth rejection, introspection disabled) used to save a connection that looked healthy but had zero tools, with no error anywhere. Reported by a user trying to connect the Linear GraphQL API with an API key.

Introspection failures are now classified (auth rejected, not a GraphQL schema, unreachable, GraphQL errors) and surfaced: the key-first connect flow blocks before saving with an actionable message, unhealthy connections show the verdict in the accounts list and on the empty tools page with a re-check button, and tool sync returns `incomplete` instead of persisting an authoritative empty catalog.

Invalid endpoint rejected at connect time:

GraphQL · invalid introspection format shows an actionable connection error (selfhost)

Valid endpoint still connects and generates tools:

GraphQL · valid introspection connects healthy and generates tools (selfhost)

### DIFF
--- a/e2e/scenarios/graphql-invalid-format-no-tools.test.ts
+++ b/e2e/scenarios/graphql-invalid-format-no-tools.test.ts
@@ -1,0 +1,152 @@
+// Cross-target (browser): GraphQL source discovery can fail silently at
+// connection-create time. The UI journey is the reported shape: add a GraphQL
+// integration, connect an API key, see the connection saved, then land on an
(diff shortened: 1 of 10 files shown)


RhysSullivan:
Closing in favor of #1483

PR #1373, opened 2026-07-08
Show a loading shell in the add-connection modal while auth methods resolve

RhysSullivan:
Opening the add-connection modal before the integration catalog resolved rendered a half-initialized form: an empty auth-method strip, a bare credential step, and a live Continue button, which then flipped to the real methods once the response landed. The modal now takes a `methodsLoading` signal from its callers and shows a skeleton shell until methods are resolved, so the form only ever mounts with real data.

Connections · the add modal holds a loading shell until auth methods resolve (cloud)


RhysSullivan:

PR #1399, opened 2026-07-10
Request uncompressed responses from remote MCP servers

RhysSullivan:
Connecting to some remote MCP servers failed with "Failed connecting via sse". The server returns gzipped responses with a miscomputed Content-Length (the uncompressed size), so the stream is truncated and the streamable-http handshake fails with a decompression error; the auto-transport SSE fallback then fails too (the server never supported legacy SSE) and its error masked the real one.

This sends `accept-encoding: identity` on remote MCP transports and the shape probe, matching what other MCP clients do (a caller-supplied Accept-Encoding header still wins), and includes both transport failures in the error message when the auto fallback also fails.


RhysSullivan:

PR #1441, opened 2026-07-20
feat(selfhost): allow additional trusted browser origins

### LINKED ISSUE #1440: Support additional trusted browser origins for self-hosted instances
## Problem

A self-hosted instance can intentionally be reachable through one canonical
public HTTPS origin and additional private-network browser aliases.

`EXECUTOR_WEB_BASE_URL` needs to remain canonical because it drives OAuth
callbacks, MCP metadata, approval links, and generated URLs. Better Auth
currently trusts only that origin, so browser authentication from an intentional
alias fails. Changing the base URL to an alias would break the canonical outbound
URL contract.

## Proposed behavior

Add an optional comma-separated environment variable:

```env
EXECUTOR_TRUSTED_ORIGINS=http://executor.home.arpa:4788,http://192.0.2.10:4788
```

The canonical base can remain `https://executor.example.com`.

- Accept exact HTTP and HTTPS origins only.
- Trim, normalize, and deduplicate configured origins.
- Always trust the canonical base origin.
- Fail at startup for invalid configuration.
- Do not infer trusted origins from `Host` or forwarded-host headers.
- Keep OAuth callbacks and generated URLs on the canonical base origin.

## Security

This should be an explicit operator-controlled allowlist, with no wildcards or
request-derived values. Existing single-origin behavior should remain unchanged
when the variable is unset.

## Use cases

- NAS and home-server dashboards
- split-horizon DNS
- VPN aliases
- reverse proxies with a canonical public endpoint plus local access

salmonumbrella:
Closes #1440

## Summary

- add a validated `EXECUTOR_TRUSTED_ORIGINS` allowlist for self-hosted browser aliases
- keep `EXECUTOR_WEB_BASE_URL` canonical for OAuth callbacks and generated URLs
- improve invalid-origin guidance and document the new setting

## Security

- accepts exact, explicit HTTP and HTTPS origins only
- rejects credentials, paths, queries, fragments, wildcards, and other schemes
- never infers trusted origins from request headers
- preserves the existing single-origin default when unset

## Verification

- `bun run format`
- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run --cwd apps/host-selfhost test` (18 files, 80 tests)


salmonumbrella:
Some context on why the mixed-protocol example (https canonical + http trusted origins) is the headline use case: umbrelOS — and home servers like it — serve local access (`umbrel.local`, LAN IP) over plain HTTP only. HTTPS for local access has been an open umbrelOS request since 2021 (getumbrel/umbrel#546), while remote access via Tailscale is HTTPS. So a self-hosted instance on that class of host unavoidably spans one https origin plus http local aliases, and a single canonical `EXECUTOR_WEB_BASE_URL` can't cover them.

PR #1448, opened 2026-07-21
Support OAuth client_secret_basic

### LINKED ISSUE #1447: Support client_secret_basic for OAuth client credentials
## Problem

Executor's low-level OAuth helper supports both `client_secret_post` and `client_secret_basic`, but configured OAuth clients do not persist or expose the token-endpoint authentication method. The service-level client-credentials mint and expiry re-mint therefore omit `clientAuth` and always use the default body transport.

This blocks authorization servers that require HTTP Basic client authentication at the token endpoint. A concrete example is Aikido:

```http
POST https://app.aikido.dev/api/oauth/token
Authorization: Basic base64(CLIENT_ID:CLIENT_SECRET)
Content-Type: application/x-www-form-urlencoded

grant_type=client_credentials
```

Executor currently sends `client_id` and `client_secret` in the form body instead, which Aikido rejects.

A previous broad implementation in #1217 included a per-client `token_endpoint_auth_method` and covered initial mint plus expiry re-mint, but that PR was closed unmerged.

## Expected behavior

- OAuth clients can select `client_secret_basic` or `client_secret_post`.
- The selection is persisted.
- Both the initial `client_credentials` exchange and subsequent expiry re-mints use the selected method.
- Existing clients remain compatible by defaulting to `client_secret_post`.
- The configuration UI/API exposes the choice, ideally using the standard OAuth names.

## References

- RFC 6749 §2.3.1: https://www.rfc-editor.org/rfc/rfc6749.html#section-2.3.1
- Aikido token endpoint: https://apidocs.aikido.dev/reference/getaccesstoken
- Prior closed implementation: #1217

MisterJimson:
## Summary

Add configurable OAuth token-endpoint client authentication so registered clients can use either `client_secret_post` or `client_secret_basic`. The selected method is persisted and reused across authorization-code exchange, client-credentials minting, refresh, and re-minting.

`client_secret_basic` uses the interoperable HTTP Basic representation expected by common providers: Base64 of the literal UTF-8 `client_id:client_secret` pair.

Closes #1447

## Changes

- Persist the token-endpoint authentication method on OAuth clients in PostgreSQL, SQLite, and Cloudflare D1.
- Expose `client_secret_post` and `client_secret_basic` through the OAuth client API and registration form.
- Thread the selected method through authorization-code exchange, client-credentials minting, refresh, and client-credentials re-minting.
- Preserve `client_secret_post` as the default for existing and first-party clients.
- Encode `client_secret_basic` as the literal credential pair instead of form-encoding each component before Base64 encoding.
- Add regression coverage for Basic credentials containing underscores and punctuation.

## Interoperability example: Aikido

[Aikido's OAuth documentation](https://apidocs.aikido.dev/reference/authorization) requires client credentials in an HTTP Basic authorization header for its [token endpoint](https://apidocs.aikido.dev/reference/getaccesstoken).

The strict RFC 6749 representation form-encodes each credential component before Base64 encoding. For example, an Aikido client ID beginning with `AIK_CLIENT_` is transmitted as `AIK%5FCLIENT%5F...`. Aikido compares the decoded Basic username and password literally, so it rejects that representation as invalid credentials even though the same pair succeeds with conventional Basic clients such as `curl --user`.

Using Base64 of the literal credential pair makes this real provider work while retaining the expected `client_secret_basic` request shape. See [RFC 6749 §2.3.1](https://www.rfc-editor.org/rfc/rfc6749#section-2.3.1) for the strict encoding rule and the note that HTTP Basic is not otherwise recommended outside this client-authentication profile.

## Testing

Validated after rebasing onto Executor v1.5.42 plus current `main`:

- [x] `bun run format:check`
- [x] `bun run lint`
- [x] `bun run typecheck` (44/44 packages)
- [x] OAuth helper, flow, and client-list tests (102 tests)
- [x] Cloudflare data-migration tests (5 tests)
- [x] All five tests that timed out under full-suite parallel load rerun successfully at low load (71 tests)
- [x] Production Cloudflare build
- [x] Manual client-credentials connection against Aikido
- [x] Production deployment using the existing D1 database and encrypted credential store

The repository-wide test run reached five unrelated 5-second timeouts under parallel load; all five affected files passed immediately when rerun together at lower load. There were no assertion failures.

## Notes for reviewers

The Basic encoding is intentionally interoperable rather than RFC-strict. Exposing separate “strict” and “raw” Basic choices would require users to understand a wire-level distinction that OAuth provider documentation generally does not surface. If a provider requiring strict component encoding is identified, that can be added later as an advanced compatibility mode with a concrete test case.

### DIFF
--- a/.changeset/oauth-client-basic-auth.md
+++ b/.changeset/oauth-client-basic-auth.md
@@ -2,4 +2,4 @@
 "@executor-js/sdk": patch
 ---
 
-Allow registered OAuth clients to use `client_secret_basic` at the token endpoint. The selected client-auth method is persisted and reused for authorization-code exchanges, client-credentials mints, token refreshes, and client-credentials re-mints; existing clients continue to default to `client_secret_post`.
+Allow registered OAuth clients to use interoperable `client_secret_basic` at the token endpoint. Basic credentials preserve the literal client ID and secret expected by common providers. The selected client-auth method is persisted and reused for authorization-code exchanges, client-credentials mints, token refreshes, and client-credentials re-mints; existing clients continue to default to `client_secret_post`.
--- a/.changeset/oauth-client-basic-auth.md
+++ b/.changeset/oauth-client-basic-auth.md
@@ -1,0 +1,5 @@
+---
+"@executor-js/sdk": patch
+---
+
+Allow registered OAuth clients to use `client_secret_basic` at the token endpoint. The selected client-auth method is persisted and reused for authorization-code exchanges, client-credentials mints, token refreshes, and client-credentials re-mints; existing clients continue to default to `client_secret_post`.
--- a/apps/cloud/drizzle/0016_slimy_amphibian.sql
+++ b/apps/cloud/drizzle/0016_slimy_amphibian.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "oauth_client" ADD COLUMN "token_endpoint_auth_method" text;
--- a/apps/cloud/drizzle/meta/0016_snapshot.json
+++ b/apps/cloud/drizzle/meta/0016_snapshot.json
@@ -1,0 +1,1492 @@
+{
+  "id": "aec7993e-87f4-40e6-87df-12fe915bf8ad",
+  "prevId": "d666b31a-c3d1-4bd7-9bd6-85f2abc4fb55",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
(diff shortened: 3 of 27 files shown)


mhodgson:
@MisterJimson have you received any indication of an intention to get this merged? Or are you just running your fork instead?

MisterJimson:
I've posted on the discord, running my fork until it gets merged.

PR #1459, opened 2026-07-26
fix(self-host): preserve native MCP elicitation mode

wan0net:
## Summary

- preserve the parsed `elicitation_mode` for non-browser in-memory MCP sessions
- keep the existing browser approval wiring unchanged
- add a self-host regression test and patch changeset

## Why

`readElicitationMode` accepts `native`, and the local and Cloudflare hosts pass that value through. The in-memory session store used by self-host instead mapped every non-browser mode to `model`, so `?elicitation_mode=native` exposed the model-resume flow and prevented native MCP elicitation from reaching capable clients.

## Validation

- `bun run format:check`
- `bun run lint`
- `bun run typecheck` (43 packages)
- `bun run test` (full non-e2e repository suite)
- `bun run --cwd packages/hosts/mcp test` (65 tests passed, including the new regression)
- branch is current with upstream `main`

### DIFF
--- a/.changeset/fuzzy-dragons-elicit.md
+++ b/.changeset/fuzzy-dragons-elicit.md
@@ -1,0 +1,5 @@
+---
+"executor": patch
+---
+
+Preserve `elicitation_mode=native` when creating self-hosted MCP sessions.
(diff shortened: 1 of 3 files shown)


RhysSullivan:
thanks!

PR #1463, opened 2026-07-27
Add tenant-scoped subject table to core schema

RhysSullivan:
Schema-only first step of the data-model work: a tenant-scoped `subject` table (external_id, created_at, last_seen_at, status) in `coreTables`, the cloud mirror, and drizzle migration `0012`. No readers or writers yet; those land in the next PR.

Note: the generated migration originally included a spurious `definition.name` narrowing from pre-existing schema drift (schema file regenerated to varchar(255) after dd7aaf601 widened prod to text); stripped here and the snapshot records prod's real shape.

### [Stack](https://github.com/kitlangton/stack)

1. **#1463** 👈 current
2. #1464
3. #1465
4. #1466
5. #1467
6. #1469

PR #1464, opened 2026-07-27
Populate subject table at request seam and connection create

RhysSullivan:
Second step of the data-model work, stacked on #1463. Adds `touchSubject` — a throttled (1h), non-fatal upsert recording (tenant, external_id, created_at, last_seen_at) — called from `makeScopedExecutor` on every request and from connection-create. Null subjects skip; org purge covers the new table. Writers only; admin reads land next.

1. #1463
2. **#1464** 👈 current
3. #1465
4. #1466
5. #1467
6. #1469


RhysSullivan:
Consolidated into #1469.

PR #1465, opened 2026-07-28
Add tenant-reach platform view with internal admin reads

RhysSullivan:
Third step, stacked on #1464. Adds the platform view: a read-only `reach: "tenant"` mode on the owner policy (writes reject under tenant reach — including tenant-wide update/delete row filters, which would otherwise widen), an opt-in `executor.admin` surface (listSubjects / listSubjectConnections / listSubjectsWithConnections, credential-bearing fields excluded), and cloud groundwork for org-scoped API keys resolving to a `PlatformAuth` distinct from `Principal`. Org keys are rejected on product and MCP surfaces. Public /admin/users endpoints land next.

1. #1463
2. #1464
3. **#1465** 👈 current
4. #1466
5. #1467
6. #1469


RhysSullivan:
Consolidated into #1469.

PR #1466, opened 2026-07-28
Add /connect/:integrationSlug deep links

RhysSullivan:
Stacked on #1465 (shares the branch point; no dependency on its code). Adds a slug-based connect deep link to the shared console routes: resolves against the tenant catalog and redirects into the existing add-account flow (`?tab=accounts&addAccount=1`). Survives sign-in on cloud (reserves `connect` from org-slug parsing) and self-host (login now returns to the live location instead of hardcoded `/`). e2e scenario green on cloud and selfhost targets.

1. #1463
2. #1464
3. #1465
4. **#1466** 👈 current
5. #1467
6. #1469


RhysSullivan:
Consolidated into #1469.

PR #1467, opened 2026-07-28
Add public admin users API and org API keys

RhysSullivan:
Stacked on #1465. Public admin surface on the platform view, mounted like the Account API with per-host auth: `GET /admin/users`, `GET /admin/users/with-connections`, `GET /admin/users/:externalId/connections` — credential fields excluded by projection (direct assertion, mutation-tested). Gated by org-scoped API key (cloud, minted via `POST /api/account/org-api-keys`, admin-only) or admin membership (cloud admin role / self-host Better Auth owner-admin). Platform executors skip touchSubject so org keys never appear as users. Known gap: the WorkOS emulator lacks org-key minting, so e2e drives the admin-session path; org keys covered at unit level (upstream emulate change filed as follow-up).

1. #1463
2. #1464
3. #1465
4. #1466
5. **#1467** 👈 current
6. #1469

### DIFF
--- a/apps/cloud/src/account/workos-account-service.ts
+++ b/apps/cloud/src/account/workos-account-service.ts
@@ -246,6 +246,32 @@ export const workosAccountProvider: Layer.Layer<
           return { success: true };
         }),
 
+      // Org keys read every user in the tenant through the `/admin/*` plane, so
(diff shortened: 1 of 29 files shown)


RhysSullivan:
Consolidated into #1469.

PR #1490, opened 2026-07-29
probe: disable request-seam subject touch

RhysSullivan:
Temporary diagnostic branch. Isolates whether the request-seam subject touch causes the cloud shard 4 e2e failures. Not for merge.


RhysSullivan:
Diagnostic probe complete. Confirmed the request-seam subject sighting was the cause of the cloud e2e shard failures; the fix landed on #1464.

PR #1510, opened 2026-07-31
Add opt-in Alchemy deployment for the Cloudflare host

magoz:
## What

Adds an opt-in Alchemy v2 deployment path for the existing Cloudflare host while keeping Wrangler as the default:

- introduces a separate `host-cloudflare-alchemy` workspace so Alchemy can use its required Effect version without upgrading Executor's runtime Effect dependency;
- deploys the existing async Worker entry, D1/R2 resources, SQLite Durable Object classes, assets, variables, and redacted encryption key with the current production names and bindings;
- adds isolated `validation` commands, adoption-aware production commands, shared remote-state bootstrap/authentication helpers, and explicit adoption safeguards;
- supplies a deterministic content hash for built assets so fresh-build plans converge instead of conservatively updating the Worker every time;
- exports the Cloudflare Worker through the package boundary and documents fresh deployment, validation, CI authentication, adoption, and the `EXECUTOR_SECRET_KEY`/`keep_bindings` risks.

Cloudflare Access remains an external prerequisite with the pinned Alchemy release. Managed OAuth Access application support is being added upstream in https://github.com/alchemy-run/alchemy/pull/1023.

## Verification

- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run --cwd apps/host-cloudflare test` (23 tests)
- `bun run --cwd apps/host-cloudflare build`
- two fresh builds produced the same asset-directory hash
- post-deploy `plan:alchemy:validation`: D1, R2, and Worker all no-op
- isolated live deployment behind Cloudflare Access: authenticated UI/account API, QuickJS execution, D1/R2-backed OpenAPI lifecycle, MCP initialization/tool execution/termination, cross-session Durable Object pause/resume, and OAuth discovery/challenge behavior

`TURBO_TEST_CONCURRENCY=4 bun run test` exits successfully. Its unchanged `apps/cloud/src/account/org-api-key-revoke.node.test.ts` currently reports three pre-existing assertions caused by calls that omit the required organization selector header; the scoped Cloudflare host suite is green.

### DIFF
--- a/apps/host-cloudflare-alchemy/.env.example
+++ b/apps/host-cloudflare-alchemy/.env.example
@@ -1,0 +1,20 @@
+# Optional for non-interactive/CI deployments. For local use, prefer the
+# Alchemy profile created by `bun run auth` and leave these unset.
+# CLOUDFLARE_ACCOUNT_ID=
+# CLOUDFLARE_API_TOKEN=
+
+# Existing Cloudflare Access application configuration.
+ACCESS_TEAM_DOMAIN=your-team.cloudflareaccess.com
+ACCESS_AUD=
+ADMIN_EMAILS=admin@example.com
+
+# Generate once with `openssl rand -hex 32`, then preserve it for the lifetime
+# of this D1 database. Changing it makes stored encrypted credentials unreadable.
+EXECUTOR_SECRET_KEY=
+
+# Optional Worker configuration. Omit entries that are not needed.
+# SELF_HOSTED_ORG_ID=default
+# SELF_HOSTED_ORG_NAME=Default
+# SELF_HOSTED_ORG_SLUG=default
+# ALLOW_LOCAL_NETWORK=false
+# VITE_PUBLIC_SITE_URL=https://executor-cloudflare.example.workers.dev
--- a/apps/host-cloudflare-alchemy/alchemy.run.ts
+++ b/apps/host-cloudflare-alchemy/alchemy.run.ts
@@ -2,11 +2,13 @@ import * as Alchemy from "alchemy";
 import * as Cloudflare from "alchemy/Cloudflare";
 import * as Config from "effect/Config";
 import * as ConfigProvider from "effect/ConfigProvider";
+import * as Data from "effect/Data";
 import * as Effect from "effect/Effect";
+import * as FileSystem from "effect/FileSystem";
 import * as Option from "effect/Option";
+import * as Path from "effect/Path";
+import type { PlatformError } from "effect/PlatformError";
 import { createHash } from "node:crypto";
-import { readdir, readFile } from "node:fs/promises";
-import { join, relative, sep } from "node:path";
 import { fileURLToPath } from "node:url";
 
 const HOST_PACKAGE_URL = import.meta.resolve("@executor-js/host-cloudflare/package.json");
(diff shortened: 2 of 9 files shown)


apps/host-cloudflare-alchemy/.env.example
```diff
@@ -0,0 +1,20 @@
...
+# CLOUDFLARE_API_TOKEN=
+
+# Existing Cloudflare Access application configuration.
+ACCESS_TEAM_DOMAIN=your-team.cloudflareaccess.com
+ACCESS_AUD=
+ADMIN_EMAILS=admin@example.com
+
+# Generate once with `openssl rand -hex 32`, then preserve it for the lifetime
+# of this D1 database. Changing it makes stored encrypted credentials unreadable.
+EXECUTOR_SECRET_KEY=
```
BlankParticle, line 13: you can use `Alchemy.Random` btw
https://alchemy.run/cloudflare/security/secrets-env/#generate-tokens-with-alchemyrandom

apps/host-cloudflare-alchemy/alchemy.run.ts
```diff
@@ -0,0 +1,195 @@
...
+        flags: ["nodejs_compat"],
+      },
+      observability: { enabled: true },
+      // Alchemy emits the required `ASSETS` fetcher binding whenever an assets
+      // config is present; the binding name is fixed by the Worker provider.
+      // Supplying a deterministic content hash also lets unchanged plans
+      // converge instead of conservatively treating the directory as dirty.
+      assets: {
+        directory: HOST_ASSETS_DIR,
+        hash: assetsHash,
```
BlankParticle, line 157: you shouldn't need to do your own hashing, assets should be doing that already

Image

apps/host-cloudflare-alchemy/.env.example
magoz: Thanks! Alchemy.Random makes sense for a fresh database. This path also supports adopting an existing Wrangler deployment, though, where we must preserve the exact existing EXECUTOR_SECRET_KEY. Otherwise credentials already encrypted in D1 become unreadable.

In beta.66, Random seem to not be able to ingest ingest/adopt an existing value and --adopt cannot recover the Worker secret plaintext, so I kept this as an explicit Config.redacted input.

https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.66/packages/alchemy/src/Random.ts


But maybe I'm missing something.

apps/host-cloudflare-alchemy/alchemy.run.ts
magoz: I found that beta.66 hashes the assets when applying the Worker update, but the alchemy plan diff doesn’t hash the directory. Without supplying a hash, every plan shows the Worker as changed even when dist is unchanged. That’s why I added it, but I agree the custom hashing isn’t ideal. Would using Alchemy.Command.Build and passing its output hash be the intended approach, or would you rather I remove it and accept the conservative plans?

- [Asset files and manifest are hashed in readAssets](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.66/packages/alchemy/src/Cloudflare/Workers/Assets.ts#L156-L253)
- [The plan/diff deliberately avoids reading the directory without a supplied hash](https://github.com/alchemy-run/alchemy/blob/v2.0.0-beta.66/packages/alchemy/src/Cloudflare/Workers/WorkerProvider.ts#L3863-L3888)

apps/host-cloudflare-alchemy/.env.example
BlankParticle: ah makes sense, then this one is fine

apps/host-cloudflare-alchemy/alchemy.run.ts
BlankParticle: that sounds like a bug, please make an issue on alchemy side, this is fine for now, if this PR gets merged without the fix being released, we can open a follow up PR later

sam-goodwin: Does this help? https://github.com/alchemy-run/alchemy/pull/1045

magoz: Thanks, makes sense. I opened https://github.com/alchemy-run/alchemy/issues/1047 and will leave the workaround as-is for now. If you have any other feedback on the PR, I’m happy to address it.

magoz: @sam-goodwin  Yep, this fixes it! I tested the patch against this stack: unchanged assets now plan as a no-op, and changing an asset correctly plans a Worker update. I’ll remove the workaround once the fix is released. Thanks!

> Does this help? [alchemy-run/alchemy#1045](https://github.com/alchemy-run/alchemy/pull/1045)

PR #1530, opened 2026-08-04
Fix OpenAPI multipart file field uploads

### LINKED ISSUE #1401: OpenAPI multipart file fields are serialized as plain form fields
## Summary

Executor appears to mishandle file inputs for OpenAPI `multipart/form-data` tools. The likely bug is OpenAPI file-part coercion: a file field is being sent as a normal form field instead of a real multipart file part with bytes/filename/content type.

## Repro

1. Use Executor `v1.5.32` with an OpenAPI source that has a multipart file upload operation.
2. For a concrete repro, use Paperless-ngx `POST /api/documents/post_document/`
4. Call the tool with the PDF as the `document` field, trying either a `ToolFile`-shaped value or byte-array/file-path style CLI input.

## Expected

Executor sends `document` as a real multipart file part containing the PDF bytes, filename, and `application/pdf` content type. Paperless accepts the upload and creates a document.

## Actual

Paperless returns HTTP 400:

```json
{"document":["The submitted data was not a file. Check the encoding type on the form."]}
```

## Investigation Notes

- Tool schema currently exposes `document: string` in `PostDocumentRequest`, despite the endpoint requiring a file upload.
- Tried `ToolFile`, byte-array payload, and `@/tmp/file.pdf` style CLI input; all produced the same Paperless validation error.
- This likely belongs in the OpenAPI request-body / multipart serialization path, not in the Paperless integration specifically.
- Relevant area to inspect: OpenAPI operation binding/schema generation for file fields plus runtime multipart coercion (`coerceFormDataRecord` / request body dispatch). The generated tool may be losing file semantics before runtime has a chance to build the correct multipart part.

BittuBarnwal7479:
## Summary

Fix OpenAPI multipart `format: binary` fields so they are exposed as `ToolFile` inputs and sent as real multipart file parts instead of plain form fields.

Partially addresses #1401 — inline multipart schemas only. $ref-referenced request bodies (including the reported Paperless case) still keep their declared binary strings; see the changeset for why, and a $defs-aware rewrite is the follow-up.

## Changes

- Map multipart binary string fields to the existing `ToolFile` input schema.
- Convert `ToolFile` values into multipart file parts with bytes, filename, and content type.
- Add a regression test for a Paperless-style PDF upload field.

## Tests

- `bun run --cwd packages/plugins/openapi test non-json-body.test.ts`
- `bun run --cwd packages/plugins/openapi typecheck`

### DIFF
--- a/.changeset/openapi-multipart-file-fields.md
+++ b/.changeset/openapi-multipart-file-fields.md
@@ -1,0 +1,5 @@
+---
+"@executor-js/plugin-openapi": patch
+---
+
+Multipart file fields in an OpenAPI spec now accept and send real files. A `multipart/form-data` property typed as a binary or byte string is rewritten into the SDK's tool-file schema when the tool is extracted, so an agent supplies a file the same way it does everywhere else. On invocation those values are decoded back into `File`/`Blob` parts — as bare properties and inside arrays — instead of being JSON-stringified into the form body, which is what upstreams were previously rejecting.
--- a/.changeset/openapi-multipart-file-fields.md
+++ b/.changeset/openapi-multipart-file-fields.md
@@ -2,4 +2,11 @@
 "@executor-js/plugin-openapi": patch
 ---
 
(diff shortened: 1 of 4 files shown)


BittuBarnwal7479:
@RhysSullivan pls review.

PR #1554, opened 2026-08-07
Record every tool call in an audit log

midego1:
## The gap

Executor keeps no record of tool usage. A run that calls GitHub or Search Console leaves one HTTP line —

```
INFO http.span: Sent HTTP response { "http.method": "POST", "http.url": "/mcp", "http.status": 200 }
```

— and nothing about which integration, which tool, or what came back. There is no `executions`/`audit` table in the schema, and `packages/core/analytics` is anonymous by construction: its own header forbids tool addresses, connection names and arguments, which are exactly the fields an audit needs. The `executor.tool.execute` span carries the right data, but only where an OTel exporter is configured, and it can't answer "which connection did this agent use last week".

Two questions were therefore unanswerable after the fact: **what did this agent touch**, and **what did my policies actually stop**.

## What this adds

`execute` in `packages/core/sdk/src/executor.ts` is the one place every call passes through — every plugin kind, every host, MCP and REST alike — so the row is written there, from an `Effect.onExit` wrapper that sees every way a call can end:

| outcome | meaning |
|---|---|
| `ok` | reached the upstream and succeeded |
| `fail` | the tool's own error result — which rides the **success** channel by design, so a channel-only reading records an upstream 404 as a healthy call |
| `blocked` | a `block` policy stopped it |
| `declined` | a human refused the approval |
| `error` | tool/connection missing, plugin not loaded, transport broke |

`blocked` and `declined` are the rows that make this worth having: both end **before any request is made**, so nothing upstream ever saw them and no HTTP-level observation can.

Each row carries the address as called, its integration/connection/tool, the governing policy (action + pattern), the duration, and the top-level argument names.

Readable three ways:

- `executor.toolCalls.list({ integration, connection, outcome, client, since, search, limit, offset })`
- `GET /api/tool-calls` with the same filters
- the **Activity** page in the console

## The Activity page

Activity overview

Filtering follows the console's existing patterns: outcome as `FilterTabs`, an integration dropdown drawn with the same favicon pipeline as the Integrations page, and a search box over the tool address (DB-level `contains`, deferred input). Every filter change resets to page 1.

Integration dropdown

Pagination is the Admin · Users pattern: 25 rows per page, one extra row fetched to answer "is there a next page" without a count query, Previous/Next with the mono page label. `offset` runs through all three layers, bounded at the HTTP edge like every other numeric input.

The blocked view — the row that exists nowhere else, because the call ended before any request was made:

Blocked filter

Read-only by construction: a log a caller can edit is not evidence, so there is no write or delete endpoint. `executor.toolCalls.prune({ before })` exists for retention, and nothing schedules it — an audit log that silently deletes itself on a default nobody chose seemed worse than one that grows.

## Who ran it, and which client

A row said which member's scope a call ran under, but not which agent made it — so on a single-user instance every call read as the same person, and a call on an **org** connection (owned by the org, `subject ""`) named nobody at all.

Rows now carry both:

- `actor` / `actor_label` — the member who ran the call, kept even when the row files under the org tier.
- `client_kind` / `client_id` / `client_name` — the credential it arrived on: which API key, which OAuth-connected MCP client (by its registered name), the browser console, or a CLI login.

Only the host sees the credential, so the host names it: a new optional `ExecutorConfig.caller`, carried per request by a `RequestCaller` service read optionally — the same shape as the existing `RequestOrgSlug`, so a host that cannot tell (cloud's WorkOS path, the CLI, tests) simply records the member and leaves the client null. Self-host resolves it from whichever credential authenticated the request: the API key (the api-key plugin mints its session with `session.id` set to the key's id), the MCP OAuth client (the access-token row names it), a CLI bearer session, or the cookie.

All five columns are nullable, so a host that creates tables at boot adds them to an existing table in place; rows written before this simply have no client.

A client names itself at OAuth registration, so its label is untrusted text: control characters are stripped and the length is bounded before storage. `GET /api/tool-calls/clients` lists the clients seen, and Activity gains a **Client** column and a matching filter.

Activity with the client column

## What it deliberately does not store

Arguments, results, and any text that came from outside:

- **Messages are never persisted.** A failed call keeps its `code`; plugins derive `error.message` from upstream response bodies (the OpenAPI plugin lifts it straight out), which routinely echo the request back — token included.
- **Codes must look like codes.** `ToolError.code` is typed as any string, so a plugin can forward a body into it. Anything not matching an identifier shape is dropped; the `outcome` column already says what happened.
- **Argument names, never values** — and only names that look like parameters. `execute` takes `unknown` args, so `{ "ghp_realtoken": null }` is a reachable shape; names are length-bounded, identifier-shaped, and credential-shaped ones are dropped.

## Failure behaviour

Writing a row can never change the outcome of the call it describes. The write is wrapped in `catchCause`, so an insert failure — or a defect — is logged and swallowed. It is *awaited* on purpose (a forked write would be interrupted when a per-request host tears the executor down, and a silently missing row is the one thing an audit log may not do), and deliberately carries no timeout of its own: an `onExit` finalizer runs uninterruptible, so a timeout there can never deliver its interrupt — it is decorative in exactly the sick-database case it would be written for, and its timer deadlocks the run loop under the adversarial scheduler budgets the execute-read-concurrency tests (#1867) exercise. Bounding a stalled driver is the driver's job.

## Scope and follow-ups

- Local/self-host/Cloudflare need no migration: `ensureDrizzleRuntimeSchemaFromTables` creates the table at boot from `coreTables`. Cloud gets `0021_tool_call_log.sql`.
- `apps/local/src/db/executor-schema.ts` is left alone — it already predates `artifact`/`subject`, and it only drives a generate-time baseline.
- **Known limitation:** a subject's view spans two partitions (its own rows plus the org's), so a newest-first read across both still sorts. Serving it takes a `(tenant, created_at)` index and the schema layer has no non-unique index API yet. Worth adding before this table gets large; I left a note at the table definition rather than adding a unique index whose leading columns wouldn't serve the query.
- Branch is current with `main` (last merged through the concurrent-read work of #1867; the audit wrapper is re-anchored in the new fibered `execute` path and the full sdk suite — including the adversarial scheduler tests — passes with it in place).

## Verification

- `packages/core/sdk/src/tool-call-log.test.ts` — 28 tests: outcome classification per ending (including a decline raised *inside* a handler, which arrives wrapped in `ToolInvocationError`), the redaction rules with real-looking secrets, list filtering/search/paging tiling without overlap, and executor-level assertions that a blocked call and a declined approval each leave a row.
- Attribution is covered end-to-end on the real self-host app: `apps/host-selfhost/src/tool-call-attribution.test.ts` signs in, mints an API key, and makes the same call three ways (API key, bearer session, cookie), asserting each row names the right client and that the key's secret appears nowhere in what is stored or served; `mcp/mcp-oauth.test.ts` now walks the full OAuth flow and asserts the resulting row names the OAuth client by its registered name.
- `format:check`, `lint`, `typecheck` (45/45) clean; `packages/core/sdk` 971, `packages/core/api` 132, `packages/react` 453, `apps/host-selfhost` 166 tests pass.
- Running in production on a self-hosted instance since 2026-08-08 (self-built image from this branch), attribution included since 2026-09-19: real traffic there is now filed under the OAuth clients and API keys that made it. Screenshots are from a local instance seeded with demo data only.
- Reviewed with `codex review` over two rounds; every finding from both is addressed in the branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


midego1:
Branch is current with `main` and CI is green except one job. A summary of what the first runs turned up, since two of them were real:

**Fixed here (your guards caught them):**
- `apps/cloud` org-deletion cascade — `tool_call_log` had joined the schema without joining `purgeOrganizationData`, so a deleted org would have left its audit rows behind. Now purged and seeded like every other tenant table.
- `check:routes` — `packages/app`'s committed route tree did not know `/activity`, so that app would have 404'd the page. All five trees regenerated.
- Migration renumbered to `0017` after `0016_oauth_client_token_auth` landed, and the regenerated drizzle artifacts formatted.

**One unrelated commit, 13503451**, which you may want to drop or take separately: `session-build-semaphore` measures its timeout path with `Date.now()` while the timer runs on libuv's monotonic clock, so `setTimeout(10)` regularly resolves with a delta of 9 and `expect(waitMs).toBeGreaterThanOrEqual(10)` is a coin flip. It came up tails on the second CI run here. The commit floors the reported wait at the timeout that produced it.

**The remaining red is flake, as far as I can tell.** Each run has failed a different suite, none of them touched by this branch:
- run 2: `session-build-semaphore` (fixed above)
- run 2 + 1: `e2e cloud 13of16` → `mcp-session-cap-eviction`, timing out opening its fourth MCP session
- run 4: `apps/host-cloudflare/src/worker.e2e.node.test.ts` → 503 from miniflare on the ~1MB spec case; all 11 tests in that file pass locally on this branch

I don't have re-run rights on the repo — a "Re-run failed jobs" when you next look at this would tell us whether anything real is left. Happy to split the semaphore fix out into its own PR if you'd rather keep this one to the feature.

midego1:
Follow-up with harder evidence on the red checks, since "it's flake" deserves more than my word for it.

**`E2E (cloud 13of16)` is failing on `main` too**, on the same `mcp-session-cap-eviction` scenario, independent of this PR:

- main@e8590a0 — https://github.com/UsefulSoftwareCo/executor/actions/runs/33375809482
- main@c27436d — https://github.com/UsefulSoftwareCo/executor/actions/runs/33374983953
- main@1b78f7c — https://github.com/UsefulSoftwareCo/executor/actions/runs/33372831994
- main@fff7ed6 — https://github.com/UsefulSoftwareCo/executor/actions/runs/33295031167

That is 4 of the last 6 CI runs on `main`, each red on that one job.

**The `Test` failure on the latest run here** is `apps/host-cloudflare/src/worker.e2e.node.test.ts` → "adds a LARGE OpenAPI source", `expected 503 to be 200`, in 537ms. Two things argue against it being this branch:

- It passes 5/5 locally on this branch, whole file green (11/11).
- The sibling case in the same file, "adds an OpenAPI source and reads it back (D1 write + read path)", passes in the same CI run. If the new `tool_call_log` table broke the D1 schema-ensure, that one would fail first — the audit write itself is on `execute`, which spec-adding never touches.

A fast 503 out of workerd on the heaviest case in the file reads as the runner giving out rather than the code.

Nothing further is pending from my side: the branch is current with `main`, and the two real problems your CI caught (the missing org purge and the stale route tree) are fixed. A re-run whenever you next look at this is the only thing that would tell us more.

PR #1558, opened 2026-08-08
Fix the local e2e suite and guard it in CI

RhysSullivan:
Every failing scenario in the `local` e2e project, plus the CI gap that let them rot.

## What was broken

**1. Toolkit-scoped MCP returned 500 on every request** — a real product bug, not a test problem.

`POST /mcp/toolkits/<slug>` answered `-32603 Internal server error`. `createConfigForResource` built the toolkit session with `createExecutorHandle`, which opens the local data dir — but the running server already holds that dir's EXCLUSIVE ownership lock, so the second open failed against our own lock (`Failed to open local SQLite data`). Reproducible outside the harness with plain curl; the unscoped `/mcp` endpoint was fine.

`createExecutorHandle` now accepts a `borrowedDb`, and the toolkit path passes the running server's handle. A toolkit session differs from the default one only in its plugin set, so sharing the handle is what it always wanted. The borrowed handle is not closed when the derived executor disposes.

This has been broken since the data-dir lock landed (#1175, Jun 28), two days after the toolkit MCP path (Jun 26).

**2. `cli-mcp-daemon-attach-stress` threw `Bun is not defined`** — 2 of 3 tests, every run. The file used `Bun.spawn`, `Bun.sleep`, and web-stream readers, but vitest runs this suite under Node. Ported to `node:child_process`, matching the rest of the harness.

**3. `auth.test.ts` waited on a "built-in" label the UI stopped rendering.** Both tests timed out after 30s on `getByText("built-in")`. The failure screenshot shows the app working correctly — authenticated, integrations loaded — and the row renders name + slug; `built-in` is only an internal `kind` value. Switched to the stable `data-testid` the list already exposes.

## Why they rotted

CI only ever ran `local/stdio-mcp.test.ts`. This PR expands that step to the MCP-surface scenarios (stdio, toolkits, native elicitation, daemon attach) — the regression guards that need no browser. The browser-dependent scenarios stay out, so the pre-existing flakiness stays out too.

## Verification

All 14 files in the `local` project pass. Run in batches locally: the 6 MCP/auth scenarios, then the 5 CLI scenarios, then the remaining 3.

`typecheck`, `lint`, `format:check` clean; `apps/local` unit tests 71 passed.

## Notes

- `local/mcp-browser-approve.test.ts` failed once when run in a 6-file batch and passed on its own — boot-resource contention, the flakiness the CI comment already describes. Not addressed here.
- The `cloud` project is not verified locally (it exceeds this machine's process budget); CI covers it. The `E2E (cloud 3of4)` `cloud signIn: callback set no session (500)` failure seen on the release PR is untouched by this branch and still unexplained — it reran green there.

### DIFF
--- a/.changeset/local-toolkit-mcp-db-lock.md
+++ b/.changeset/local-toolkit-mcp-db-lock.md
@@ -1,0 +1,7 @@
+---
+"executor": patch
+---
+
+**Fix: toolkit-scoped MCP endpoints on the local server no longer fail with an internal error**
+
+`POST /mcp/toolkits/<slug>` returned `-32603 Internal server error` for every request. Building a toolkit-scoped session called `createExecutorHandle`, which opened the local data directory a second time — but the running server already holds that directory's exclusive ownership lock, so the open failed against the server's own lock ("Failed to open local SQLite data"). Toolkit sessions now borrow the running server's database handle, which is what they always needed: they differ from the default session only in their plugin set. The unscoped `/mcp` endpoint was never affected.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,14 +315,22 @@ jobs:
         working-directory: e2e
 
       # The `local` project is excluded from the default `test` chain (each
-      # scenario boots its own `executor web`). Run just the stdio MCP scenario
-      # here: it is the auto-connect / env-as-secret regression guard, and
-      # running it alone avoids the boot-resource accumulation and the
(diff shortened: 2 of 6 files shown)


RhysSullivan:
Note on CI coverage: the `E2E (local MCP)` job is gated `if: github.event_name != 'pull_request'`, so the expanded step shows as *skipped* on this PR and will first execute on push to `main` after merge. The scenarios were verified locally instead — all 14 files in the `local` project pass.

RhysSullivan:
Closing in favor of #1551

PR #1567, opened 2026-08-12
Keep token material out of OAuth token-endpoint errors

GeiserX:
## TL;DR

When a token endpoint replied in a way the OAuth library could not parse, the resulting error carried the **parsed response body** as its `cause`. On a malformed `200` that body is a *successful* token response — so an **access token, and sometimes a refresh token, travelled inside an error object** into whatever logged it.

This drops the raw `cause` and rebuilds the body preview from an **allowlist** of fields that are safe to show, instead of a denylist of fields to hide.

---

## What happens today

`OAuth2Error` is constructed with the parsed body attached. Any handler that logs the error, serialises it, or reports it to an error tracker writes token material to that sink. Nothing in the type says the payload is sensitive, so there is no reason for a caller to suspect it.

The preview that *was* redacted used a denylist — it removed the field names we thought of. A token under any other name went straight through.

## What this changes

- **`cause` is no longer attached.** The library's own parsed body never leaves the boundary.
- **The body preview is an allowlist**: `error`, `errors`, `error_description`, `error_uri`, plus `code` / `message` / `detail` *nested inside* those. A field nobody anticipated is now omitted by default rather than printed by default.

  `code` is deliberately allowed only when nested — at the top level of a token response, `code` is the RFC 6749 authorization code.
- **Form-encoded bodies take the same allowlist.** They previously had a separate, weaker path.
- **Previews are depth-bounded**, so a deeply nested or self-referential body cannot exhaust the stack.
- **The failure summary records the token endpoint's hostname, not its full URL** — paths routinely carry tenant and account identifiers.

## Why the allowlist direction matters

A denylist has to predict every name an authorization server might use for a secret; it fails silently and in the unsafe direction when it guesses wrong. An allowlist fails in the safe direction: an unfamiliar field is dropped, and the cost is a slightly less descriptive error message.

## Tests

The regression tests use the inputs that **actually trigger** the malformed-`200` path — `expires_in: null`, an array `scope`, a non-string `token_type` — rather than an input that merely looks similar. An earlier version of this test used an unknown `token_type`, which produces a different, harmless error shape and would have passed against the unfixed code.

Each test was mutation-checked: the fix was reverted and the test confirmed to fail.

## Scope

Independent of #1564 — it touches only `oauth-helpers.ts` and applies to the existing host-side exchange. The two can merge in either order.


GeiserX:

PR #1568, opened 2026-08-12
Delete the credential a connection minted when the connection is removed

GeiserX:
## TL;DR

`connections.remove` deleted the connection row and **left the credential it had minted in the store**. The secret outlived the only thing that referenced it, with no surface left in the product to see or remove it.

A user who disconnects an account still has that account's tokens held on their behalf.

---

## What happens today

Connecting an account mints credential items and records a connection row pointing at them. Removing the connection deletes only the row. The items stay — unreferenced, invisible in the UI, and indefinitely live.

For an OAuth connection those items are an access token and a refresh token. A refresh token is long-lived by design, so "disconnected" and "revoked" quietly mean different things.

## What this changes

Removal now also deletes the items the connection minted.

- **Items are identified by rebuilding their deterministic ids** from the connection row, then matching by byte equality — not by scanning for anything that *looks* related. A scan would risk deleting a credential that merely resembles the connection's own.
- **A minted credential that another connection still points at is kept.** Two connections can legitimately share one item.
- **Both halves are covered** — the OAuth items and the static ones.
- **The alias scan is scoped to the provider that owns the connection**, so an id held on a different provider is not mistaken for an alias of this one. (Not covered by a test: every connection in the test file uses a single provider, so removing that filter would not fail anything today.)

## Known limits, stated rather than implied

The alias scan can only see connections the caller is allowed to see. A connection owned by someone else that points at the same item is invisible to it. That limit is recorded in a comment next to the code, because a reader who assumes the scan is global would draw the wrong conclusion about when an item is safe to delete.

## Tests

Eleven tests covering: the OAuth path, the static path, the shared-item hold-back, and the id-matching being exact rather than fuzzy.

## Scope

Independent of #1564. It touches `executor.ts` and a doc comment on `provider.ts` (+10/-1); both files are also touched by that PR, so expect a small textual conflict if both merge — the changes are in different functions and do not interact logically.

### DIFF
--- a/.changeset/integration-removal-credential-gc.md
+++ b/.changeset/integration-removal-credential-gc.md
@@ -1,0 +1,11 @@
+---
+"executor": patch
+---
+
+**Removing an integration now removes the credentials its connections minted**
+
+`integrations.remove` deletes every connection row belonging to the integration. It left the credentials those connections had minted in the store — the same orphan `connections.remove` was fixed to prevent, reachable through a different path and stranding many secrets at once rather than one.
+
+An orphaned refresh token is the worst case: long-lived by design, no longer referenced by anything, and invisible in the product, so nobody can see it to revoke it.
+
+The rows are read before they are deleted, because once they are gone nothing names the items they minted. Only minted ids are removed — an item the connection merely referenced is left alone, exactly as on the single-connection path. The deletion is deferred until the removal commits, so a rolled-back removal leaves the credentials intact rather than restoring connections that point at secrets which no longer exist.
--- a/.changeset/orphaned-credential-on-connection-remove.md
+++ b/.changeset/orphaned-credential-on-connection-remove.md
@@ -1,0 +1,9 @@
+---
+"executor": patch
+---
+
+**Removing a connection now deletes the credential it minted**
+
+`connections.remove` deleted the connection row and left the credential it had minted in the store. The secret outlived the only thing that referenced it, with no surface left in the product to see or remove it — so a user who disconnected an account still had that account's tokens held on their behalf.
+
+Removal now also deletes the items the connection minted, identified by rebuilding their deterministic ids from the connection row rather than by scanning for anything that looks related. A minted credential that another connection still points at is kept. The alias scan is scoped to the provider that owns the connection, and both the OAuth and the static halves of the deletion are covered.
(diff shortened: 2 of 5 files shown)


GeiserX:
Pushed a fix for a defect I found in my own patch while reviewing it.

**The credential deletion was running inside the transaction that removes the connection rows.** `provider.delete` reaches outside the database — a sealed store, a keychain, someone else's API — and nothing out there enlists in that transaction or rolls back with it. If the transaction aborted, the connection row came back while its secret stayed permanently destroyed: a live connection pointing at a credential that no longer exists. That is strictly worse than the orphan this PR removes, and unlike the orphan it cannot be repaired.

**The first fix was not enough, which is the more interesting part.** Sequencing the deletion after the `transaction()` call looks correct but isn't, because `transaction()` nests by pass-through: an inner call inside an active transaction just runs its effect. So "after the inner transaction" is not after any commit at all, and a caller who wraps `remove` in their own transaction and then aborts would still have destroyed the secret.

It now routes through the existing `afterCommit` helper, which queues onto the *outermost* transaction and discards its queue on rollback.

The regression test drives exactly that path — a plugin-owned outer transaction that removes the connection and then fails — and asserts the row came back **and** the credential survived. It fails against the original code and against the incomplete first fix, so it distinguishes all three states rather than merely passing on the final one.

GeiserX:
Pushed one more commit: the same orphan exists on a second path, and fixing only the first would have been misleading.

`integrations.remove` deletes **every connection row belonging to the integration** and never touched the credentials those connections had minted. That is the identical leak this PR fixes for `connections.remove`, reached through a different door — and it strands many secrets at once rather than one. An orphaned refresh token is the worst case of it: long-lived by design, referenced by nothing, and invisible in the product, so nobody can see it in order to revoke it.

I found it while auditing which plugin lifecycle hooks run inside a transaction, not by looking for it. Worth mentioning because it means the two paths had drifted silently — merging this PR without it would leave a reviewer reasonably believing the orphan class was closed.

The fix reuses the same `deleteMintedCredentials` helper rather than reimplementing the rule, so only minted ids are removed and an item the connection merely *referenced* is left alone, exactly as on the single-connection path. The rows are read before they are deleted, since once they are gone nothing names the items they minted. The deletion is deferred past the commit for the reason in my earlier comment.

Two mutations, each killing exactly one test: removing the deletion kills the "deletes what it minted" test, and running it inline instead of deferring kills the "rolled-back removal leaves the credentials intact" test. Neither kills both — the rollback test cannot bind the deferral on its own, because with the feature absent the credential survives trivially.

GeiserX:

GeiserX:
Superseded by #2083, which redoes both halves of this change on top of the removal cascade from #1991.

PR #1569, opened 2026-08-12
Sweep expired authorization sessions so abandoned PKCE verifiers do not persist

GeiserX:
## TL;DR

An OAuth authorization session stores its **PKCE verifier** so the callback can redeem the code. Expired sessions were only discarded *lazily, on completion* — but an **abandoned flow is never completed**, so that check never ran for it and nothing else swept the table.

The verifier sat in the database in plaintext, indefinitely, for every flow a user started and walked away from.

---

## What happens today

`complete` checks whether the session has expired and discards it if so. That is the only cleanup in the table. It runs exactly once per flow, and only for flows that come back.

Users abandon OAuth flows constantly — they close the popup, deny consent, get distracted at the provider's login page. None of those sessions is ever completed, so none is ever discarded. The table grows without bound and each row holds a live secret.

## What this changes

**Starting a new authorization sweeps sessions that have already expired.**

Doing it on `start` rather than on a timer is deliberate:

- It bounds the table by **how often authorization is begun**, rather than by how often it is abandoned — the two are correlated, and the first is the one we can observe.
- It needs **no scheduler**, so it works identically in every host, including the stateless multi-isolate ones where a background job has nowhere to live.
- It runs on a path that is **already writing**, so it costs one delete rather than a new round trip.

It is **owner-scoped by the table's own delete policy**, so a caller only ever sweeps rows it can already see, and **best-effort** — failing to tidy up must never stop someone connecting an account.

Separately: **a session whose completion cannot be retried is now dropped** rather than left behind. Previously a flow that failed in a way the user would have to restart still left its verifier in place.

## What this does not claim

The sweep bounds *how long* a verifier persists; it does not stop one being stored. A verifier is only useful together with its matching authorization code, which is single-use and short-lived, so the residual value of a swept-late verifier is low. The point is that "low value" and "kept forever" is still a worse position than "low value" and "kept until it expires".

## Tests

Two tests: the sweep fires on `start` and removes only expired rows, and a non-retryable completion deletes its session. Both mutation-checked.

## Scope

Independent of #1564 — the only production file it touches is `oauth-service.ts` (plus a changeset and a new test file).


GeiserX:

PR #1570, opened 2026-08-12
Clear the OAuth popup result from localStorage after handover

GeiserX:
## TL;DR

The OAuth popup writes its result to `localStorage` as a fallback completion channel and **never removes it**. The payload carries the user's identity label — an email — and, on failure, the error preview. It stays parked in the browser profile.

This clears the entry once the handover has had time to land.

---

## Why the entry exists at all

The popup has three ways to hand its result back to the opener, because each one fails in a different situation:

- **`postMessage`** is severed when the provider's consent page sets COOP — `window.opener` becomes `null`.
- **`BroadcastChannel`** can be partitioned, and can be raced by the popup auto-closing.
- **`localStorage`** raises a `storage` event on the opener, and is the reliable fallback when the other two are gone.

So the write is doing real work and should stay.

## What this changes

The entry is removed after the handover: shortly before the window closes on success, and after a longer delay on failure, where the window stays open for the user to read the error.

**Clearing cannot cost a listener the result.** A `storage` event captures `newValue` at dispatch, so an opener that has been notified already holds the payload in its handler — removing the key afterwards does not reach back and empty it.

## Why it matters

Leaving it parks identity data in the user's browser profile indefinitely **whenever nobody is listening** — which is every abandoned flow and every opener-less flow, exactly the cases where no one will ever come back to consume it.

## Tests

The added tests assert the clear happens on both the success and the failure path, and that it is scheduled *after* the message has been dispatched rather than alongside it.

## Scope

Fully independent — one file, plus tests. It touches nothing else in this series.


GeiserX:

PR #1571, opened 2026-08-12
Restrict workspace-level settings to admins

RhysSullivan:
Members can still USE workspace resources — read them, execute tools over shared connections, and the operational writes those imply (token refresh, tool-catalog re-sync, config-rewrite healing) — but CONFIGURING them is now admin-only.

## What's gated

The executor binding gains `orgWrites: "allowed" | "denied"`, enforced at the top of every user-intent workspace-level mutation with a new `OrgWriteDeniedError` (403):

- org-owned tool policies (create/update/remove)
- workspace-shared connections (create/update/remove, org OAuth connect/reconnect)
- org OAuth apps (create/remove, incl. DCR persisting an org client)
- the integration catalog (new registration, update, remove, health check) — including the plugin extension surfaces (openapi addSpec/updateSpec/configure, mcp add/remove/configure server, graphql add/configure)

Deliberately a surface gate rather than a storage-policy axis: catalog rebuilds, legacy stdio healing, and OAuth token refresh on shared connections must keep converging under any member's binding.

## Role plumbing

- cloud: `authorizeOrganization` now surfaces the WorkOS membership role (no extra API call); HTTP requests and MCP sessions bind from it. MCP sessions bake the role at session init, so a demotion applies from the next session, and sessions persisted before this field existed keep their old behavior until re-init.
- self-host: the identity seam resolves the Better Auth org membership role (same explicit-org query as the admin gate); the MCP OAuth-bearer path reads the membership row directly. Fails closed to member.
- local / CLI / host-cf (single-tenant Access): no role model, binding stays allowed.

## Tests

- `packages/core/sdk/src/org-writes.test.ts`: denied vs allowed bindings across policies, connections, catalog, OAuth — including that members still execute tools over org connections and that the register replace-arm stays open.
- `apps/host-selfhost/src/multi-user.test.ts`: member 403s end-to-end through real Better Auth; admin setup unchanged.
- Full repo gates green (`format:check`, `lint`, `typecheck`, `test`). `apps/cloud/src/account/org-api-key-revoke.node.test.ts` fails on main already (pre-existing, untouched).

## Not in this PR

Console UI affordance (hiding the Workspace owner option and catalog mutations for non-admins) — the server now answers 403 and the console renders the error; the nav-level admin gate (`packages/react/src/lib/admin-access.ts`) already exists to build on.

### DIFF
--- a/.changeset/workspace-writes-admin-only.md
+++ b/.changeset/workspace-writes-admin-only.md
@@ -1,0 +1,21 @@
+---
+"@executor-js/sdk": minor
+"@executor-js/plugin-graphql": minor
+"@executor-js/plugin-mcp": minor
+"@executor-js/plugin-openapi": minor
+---
+
+**Workspace-level settings are now admin-only**
+
+The executor binding gains `orgWrites: "allowed" | "denied"`. Hosts derive it
+from the acting member's role (cloud: WorkOS membership role; self-host:
+Better Auth org membership role), and a plain member's binding refuses every
+user-intent workspace-level mutation with the new `OrgWriteDeniedError`
+(HTTP 403): org-owned tool policies, workspace-shared connections, org OAuth
+apps and org connect flows, and integration-catalog changes (add, update,
+remove, health check).
+
+Using workspace resources is unchanged for members: reads, tool execution over
+shared connections, and the operational writes those imply (token refresh,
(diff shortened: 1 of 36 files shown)


RhysSullivan:
Closing in favor of #1786

PR #1590, opened 2026-08-14
Align stale @modelcontextprotocol/sdk floors to ^1.29.0

RhysSullivan:
Three manifests still declared `^1.12.1` while the whole workspace resolves to 1.29.0 (agents pins it exactly, ext-apps peers on `^1.29.0`). Aligns the declared floors with reality; no resolution change (lockfile diff is manifest ranges only).

First step toward MCP spec 2026-07-28 / SDK v2 adoption.


RhysSullivan:

PR #1591, opened 2026-08-14
Executor as MCP client: speak spec 2026-07-28 outbound

RhysSullivan:
Executor as an **MCP client** on spec 2026-07-28 (part 1 of 2; the server half is #1611).

Moves `packages/plugins/mcp` from `@modelcontextprotocol/sdk` v1 to `@modelcontextprotocol/client`/`core` 2.0.0, plus the three stale `^1.12.1` manifest floors aligned to `^1.29.0`.

- Streamable HTTP connections negotiate the era automatically (`versionNegotiation: auto`); forced-SSE and stdio keep the SDK legacy default.
- The endpoint shape probe advertises 2025-11-25 and adds a `server/discover` fallback for 2026-07-28-only servers.
- Transport errors classify via v2 types (`SdkHttpError.status`, `SseError.code`, `InsufficientScopeError`) instead of matching SDK-internal message strings.
- Elicitation and tools/list_changed handlers use v2 method-string registration.
- `src/testing/` stays on SDK v1 deliberately as a legacy-wire interop fixture.

Fixes the reported failure: executor sent `2025-11-25` to a modern-only endpoint and was rejected (`Unsupported protocol version`). Verified against a `legacy: "reject"` server — the old client reproduces the report byte-for-byte; this client negotiates, lists tools, and round-trips `callTool`.


RhysSullivan:

PR #1601, opened 2026-08-16
Add SDK v2 tool-server assembly with shared registration core

RhysSullivan:
Second step of MCP spec 2026-07-28 support: the executor tool server now builds on either SDK generation.

- `tool-server-shared.ts`: SDK-agnostic core (schemas, handler bodies, artifacts, registration flow) extracted from `tool-server.ts`; the v1 assembly keeps its exact public API and behavior — every pre-existing test passes unmodified.
- `tool-server-v2.ts`: new `buildMcpServerV2` assembling the same tools on `@modelcontextprotocol/server@2` for stateless per-request serving. Native elicitation returns `input_required` with signed `requestState` (HMAC via `createRequestStateCodec`, bound to method + authenticated principal per the spec MUST, TTL-limited) riding the existing pause/resume engine; model/browser modes are unchanged logic.
- `mcp-apps.ts`: temporary wire-identical vendoring of the ext-apps server helpers (upstream v2 support still open: modelcontextprotocol/ext-apps#702); the v1 assembly keeps the real dependency.
- No host wiring yet — hosts route modern traffic to `buildMcpServerV2` in the next PRs.

Tests: 197 in `packages/hosts/mcp` (9 new v2/apps files), incl. input_required round-trips, tampered and cross-principal requestState rejection, apps-enabled/disabled variants.

Stacked on #1591.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -702,13 +702,16 @@
         "@cfworker/json-schema": "^4.1.1",
         "@executor-js/execution": "workspace:*",
         "@executor-js/sdk": "workspace:*",
+        "@modelcontextprotocol/core": "2.0.0",
         "@modelcontextprotocol/ext-apps": "^1.7.4",
         "@modelcontextprotocol/sdk": "^1.29.0",
+        "@modelcontextprotocol/server": "2.0.0",
         "effect": "catalog:",
         "zod": "4.3.6",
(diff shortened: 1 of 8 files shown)


RhysSullivan:

PR #1602, opened 2026-08-16
Serve MCP 2026-07-28 clients on selfhost and local HTTP

RhysSullivan:
Third step of MCP spec 2026-07-28 support: the neutral hosts now serve modern clients.

- `envelope.ts` routes by era after authentication: legacy requests keep the exact `McpSessionStore.dispatch` path; modern requests hit resource-keyed, process-cached `createMcpHandler(..., { legacy: "reject" })` handlers whose per-request factory reads the client capability envelope and builds the v2 server bound to the authenticated `(accountId, organizationId)` principal.
- `mcp-build.ts` gains `makeMcpBuildServerV2` mirroring the v1 execution stack; `executor-app.ts` threads the new `McpModernServerBuilder` seam next to `sessions`.
- The local app's HTTP surface gets the same routing (principal "local"); stdio deliberately stays legacy-only for now — modern stdio clients fall back via their own probe policy.
- CORS: preflights echo `Access-Control-Request-Headers` so the era's dynamic `Mcp-Param-<name>` headers pass (CORS names never glob; `*` skips credentialed requests).
- requestState signing key is per-process (both hosts are single-process, which the SDK blesses); state does not survive a host restart, matching session behavior today.

Tests: 201 hosts/mcp + 102 selfhost + 72 local; all pre-existing tests unmodified except one CORS assertion that encoded the non-working `mcp-param-*` literal.

Stacked on #1601.

### DIFF
--- a/apps/host-selfhost/package.json
+++ b/apps/host-selfhost/package.json
@@ -53,6 +53,7 @@
   "devDependencies": {
     "@effect/vitest": "catalog:",
     "@executor-js/vite-plugin": "workspace:*",
+    "@modelcontextprotocol/client": "2.0.0",
     "@tailwindcss/vite": "catalog:",
     "@tanstack/router-plugin": "^1.167.12",
     "@tanstack/virtual-file-routes": "^1.162.0",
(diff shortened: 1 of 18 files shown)


RhysSullivan:

PR #1604, opened 2026-08-16
Serve MCP 2026-07-28 clients on the Cloudflare hosts

RhysSullivan:
Final step of MCP spec 2026-07-28 serving: the Cloudflare hosts (executor-cloud and self-host-cloudflare) now speak the modern protocol. This is the endpoint modern clients hit in production.

Architecture — wire-stateless, DO-backed execution affinity:

- After bearer auth (unchanged, still before any DO addressing), the worker classifies the era. Legacy requests take the existing `McpAgent` path byte-for-byte; `agents` stays at 0.17.3 + patch.
- Modern non-`tools/call` methods are served by a resource-cached stateless v2 handler in the worker — no DO touch.
- Modern `tools/call` routes into a Durable Object for execution affinity: fresh executes get a `newUniqueId()` DO running a DO-local v2 handler over the existing execution engine; a pause registers `executionId → modern:<doId>` in the execution-owner directory before returning `input_required` with signed `requestState`.
- Continuation rounds and `resume` tool calls verify/extract the `executionId` in the worker, look up the owner directory, enforce the same cross-bearer ownership contract as legacy (403 -32003), and forward to the owning DO. Tampered or expired state degrades to a DO-less server whose codec rejects (-32602) or whose engine reports the existing `execution_not_found` shape — no new error surfaces.
- Modern DOs participate in the existing alarm/keepalive lifecycle; a DO eviction mid-pause yields the same `execution_expired` outcome legacy has today.
- CORS preflights echo requested headers (dynamic `Mcp-Param-<name>` support), telemetry records `mcp.client.protocol_version` from the modern envelope.

**Deployment prerequisite (before merge):** both workers need the shared continuation-signing secret, ≥32 bytes, visible to worker + session DOs:

```
wrangler secret put MCP_REQUEST_STATE_KEY
```

No generated fallback — the modern path fails loudly without it. Documented in both wrangler configs.

Tests: 74 hosts/cloudflare + 24 host-cloudflare + 244 cloud + 201 hosts/mcp. The 3 cloud failures in `org-api-key-revoke.node.test.ts` reproduce identically on the parent branch (pre-existing/environmental; CI arbitrates).

Stacked on #1602.

### DIFF
--- a/apps/cloud/src/env-augment.d.ts
+++ b/apps/cloud/src/env-augment.d.ts
@@ -67,6 +67,8 @@ declare global {
       MCP_RESOURCE_ORIGIN?: string;
       MCP_SESSION_TIMEOUT_MS?: string;
       MCP_PAUSED_SESSION_IDLE_TIMEOUT_MS?: string;
+      /** HMAC key for MCP 2026-07-28 continuation state (32+ byte secret). */
+      MCP_REQUEST_STATE_KEY?: string;
       NODE_ENV?: string;
 
       // Shared with frontend
--- a/apps/cloud/src/mcp/agent-handler.ts
+++ b/apps/cloud/src/mcp/agent-handler.ts
@@ -9,6 +9,7 @@ import {
   type AuthOutcome,
   type McpResource,
 } from "@executor-js/host-mcp";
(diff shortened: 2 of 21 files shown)


RhysSullivan:

PR #1606, opened 2026-08-16
Serve both MCP eras from the v2 stack on neutral hosts

RhysSullivan:
First consolidation step: the neutral hosts (selfhost + local) now serve BOTH protocol eras from the SDK v2 stack — the v1 assembly is no longer used outside the Cloudflare hosts.

- `in-memory-session-store.ts` builds sessions on the v2 `WebStandardStreamableHTTPServerTransport` + the v2 assembly (`sessionful` mode: apps capability and elicitation support follow the live initialize-negotiated state, preserving the cold-restore semantics).
- Native elicitation for legacy clients rides the SDK's built-in legacy shim (`inputRequired` returns become real `elicitation/create` requests); native sessions use SSE responses as the shim requires, model/browser sessions keep JSON.
- `apps/local` stdio moves to `serveStdio` — one process now serves legacy AND modern stdio clients (deferral removed).
- selfhost drops its v1 SDK dependency entirely; local's is dev-only (compat tests).

Interop evidence (new tests): a real v1-SDK client against the consolidated store — initialize, tools/list with negotiated Apps tools, `sessionCount() === 1` on reuse, and the full native-elicitation round-trip (`elicitation/create` emitted by the shim, response accepted, execution completes). stdio covers both a legacy v1 client and a modern-pinned v2 client.

Tests: 204 hosts/mcp + 72 local + 102 selfhost; no existing behavioral assertion changed.

Next: the cloud DO rewrite off `McpAgent` (deletes the 911-line agents patch), then removal of the v1 assembly and the V2 naming.

Stacked on #1604.

### DIFF
--- a/apps/host-selfhost/package.json
+++ b/apps/host-selfhost/package.json
@@ -40,7 +40,6 @@
     "@executor-js/sdk": "workspace:*",
     "@libsql/client": "catalog:",
     "@libsql/kysely-libsql": "catalog:",
-    "@modelcontextprotocol/sdk": "^1.29.0",
     "@tanstack/react-router": "catalog:",
     "better-auth": "^1.6.11",
     "drizzle-orm": "catalog:",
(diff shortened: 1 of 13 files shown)


RhysSullivan:

PR #1608, opened 2026-08-16
Host cloud MCP sessions on our own DO with the v2 stack

RhysSullivan:
The big consolidation step: Cloudflare MCP session serving moves off `McpAgent` onto our own Durable Object hosting the SDK v2 sessionful stack. **`patches/agents@0.17.3.patch` (911 lines) is deleted and `agents` is out of every MCP path.**

- The session DO extends `DurableObject` directly: per-session v2 `WebStandardStreamableHTTPServerTransport` + the sessionful v2 assembly, direct `stub.fetch` response streaming (the `McpAgent`/PartyServer WebSocket bridge and its delivery-ack protocol are gone — stream resumability is the event store's job).
- `do-event-store.ts`: DO-storage `EventStore` carrying the patch's semantics as first-class code — 64 events / 2 MiB per stream, never evict the newest, >120 KiB events deliver live but don't persist, best-effort writes that never break the send path, replay strictly after `Last-Event-ID`.
- `sse-response-rotation.ts`: 30-minute stream rotation with graceful reconnect framing, aligned with the alarm policy.
- Priming: v2's transport only primes 2025-11-25+ clients (verified in the installed dist); older clients on slow tool calls still need one benign id-bearing event before the result, ported minimally as our code. `e2e/cloud/mcp-priming-reconnect.test.ts` is the CI arbiter.
- Native elicitation flows through the SDK legacy shim like the neutral hosts; model/browser pause-resume, approval leases, cross-DO model resume, and the alarm policy keep their semantics (alarm arming is now fully ours).
- Session addressing keeps DO-unique-id session ids (`idFromString`); class names and wrangler bindings unchanged, so no DO migrations. **Deliberate cutover behavior:** sessions created before this deploy get one clean 404 and clients re-initialize transparently — no storage-compat restore of agents-internal state.
- Patch-internal test suites rewritten as behavior-equivalent tests of our components; a real v1-SDK client drives the DO-hosted stack end to end incl. reconnect-mid-call replay.

Tests: 57 hosts/cloudflare + 204 hosts/mcp + 24 host-cloudflare + 244 cloud (3 known-environmental failures pre-exist on parent). Key acceptance signal is CI's cloud e2e: priming reconnect, SSE replay, oversize response, session survival.

Stacked on #1606. Next PR deletes the v1 assembly and the V2 naming.


RhysSullivan:
CI verdict: 3 cloud e2e shards red — the acceptance gate working as intended. Four distinct failures:

1. `mcp-client-sessions` (unknown session id must fail fast): likely `idFromString` throwing on arbitrary client-supplied session strings where the old path returned a clean 404. Needs a parse guard in the worker.
2. `mcp-sse-replay` (severed POST response replays on the next standalone GET): the patch replayed stranded POST-stream events onto the GET stream; the stock v2 eventStore only resumes the same stream via Last-Event-ID. The DO needs the cross-stream undelivered-response replay ported.
3. `mcp-priming-reconnect` (stock old client recovers a mid-call result): the minimal priming port is not effective under real workerd — re-verify against the severed-stream flow.
4. `repro-transport-brick` (concurrent GET+POST after idle-dispose): restore/dispose race in the new lifecycle.

Iterating with the narrow e2e scenarios locally; will push the fix commit to this branch.

RhysSullivan:

PR #1614, opened 2026-08-16
Support MCP spec 2026-07-28 across client and servers

RhysSullivan:
Executor speaks MCP spec 2026-07-28 in both directions. Legacy clients are untouched on their existing code paths — this PR is purely additive routing (part 1 of 2; the follow-up consolidates legacy serving onto SDK v2 and deletes the agents patch).

- **Outbound client** (`packages/plugins/mcp`): SDK v2; streamable-HTTP connections auto-negotiate the era (`server/discover` probe, legacy fallback); typed v2 error classification; shape probe handles modern-only servers. Fixes the reported failure where executor sent `2025-11-25` to a modern-only server and was rejected.
- **Tool server** (`packages/hosts/mcp`): SDK-agnostic core with two thin assemblies — v1 byte-identical for existing paths, new v2 assembly for stateless modern serving. Native elicitation = `input_required` + HMAC-signed `requestState` (principal+method bound, TTL'd) over the existing pause/resume engine. MCP Apps helpers vendored wire-identical pending upstream ext-apps v2 (#702).
- **Neutral hosts** (selfhost, local): era-routed after auth; modern → cached stateless handlers; legacy → existing sessionful store unchanged. CORS preflights echo requested headers for dynamic `Mcp-Param-<name>`.
- **Cloudflare hosts**: modern reads worker-side (no DO); modern `tools/call` routes into DOs for execution affinity via the executionId-keyed owner directory; legacy stays on the untouched `McpAgent` path. Same cross-bearer 403 contract as legacy.

Evidence: full e2e suites green (legacy non-regression); modern support proven live with a `2026-07-28`-pinned client (discover → tools/list → execute); the pre-migration client reproduces the reported error byte-for-byte against a `legacy: "reject"` server while this client connects.

**Deployment prerequisite (before merge):** both workers need the shared continuation-signing secret (≥32 bytes), visible to worker + session DOs:

```
wrangler secret put MCP_REQUEST_STATE_KEY
```

Follow-up PR: #1611 (consolidation — deletes the 911-line agents patch, `McpAgent`, and the v1 assembly).


RhysSullivan:
Re-split by protocol role per review preference: client = #1591, server = #1611.

PR #1619, opened 2026-08-17
fix(local): reuse database for toolkit MCP sessions

### LINKED ISSUE #1618: Toolkit MCP routes fail with SQLite owner lock in local daemon
### Executor version

1.5.40. The same code path is present on 1.5.41.

### Reproduction

1. Start the local headless daemon with at least one toolkit configured.
2. Initialize the default `/mcp` endpoint. It succeeds.
3. Send the same MCP `initialize` request to `/mcp/toolkits/<slug>`.

### Expected

The toolkit-scoped MCP session initializes successfully.

### Actual

The toolkit route returns HTTP 500 and the daemon logs:

```text
Failed to open local SQLite data. Close other Executor processes and retry, or run with --log-level debug for details.
```

Only one Executor process is running. That process already owns the data-directory lock. `createConfigForResource` calls `createExecutorHandle` for the toolkit, which attempts to acquire the same local SQLite ownership lock a second time inside the daemon.

### Proposed fix

Reuse the shared daemon database when constructing a toolkit-scoped executor, while retaining the toolkit-specific plugin configuration. A regression test should initialize a toolkit MCP route in the same server process.

jamieb72-afk:
Fixes #1618.

## Summary

- construct toolkit-scoped executors over the daemon's already-owned FumaDB handle
- cache one scoped executor per MCP resource for the daemon lifetime, preventing connector processes from multiplying as clients open tasks
- preserve toolkit-specific filtering, analytics, artifacts, connections, and web base URL behavior
- add HTTP regression coverage for SQLite ownership and multi-client resource reuse/disposal
- add a patch Changeset for the published CLI

## Verification

- `bunx --bun vitest run` in `apps/local`: 76 passed
- `bun run typecheck` in `apps/local`: passed
- `bun run format:check`: passed
- `bun run lint`: passed
- `bun run typecheck`: 44 packages passed
- local arm64 build smoke: `executor v1.5.41`
- live local daemon: default MCP and both toolkit endpoints returned 200; reopening the same toolkit left the direct-child PID set unchanged at zero

The repository-wide test command reached an existing local runtime mismatch outside this change: `better-sqlite3` was compiled for Node ABI 141 while the installed Node 26 runtime requires ABI 147. The complete local app suite and both regressions pass under Bun.

### DIFF
--- a/.changeset/toolkit-mcp-shared-database.md
+++ b/.changeset/toolkit-mcp-shared-database.md
@@ -1,0 +1,5 @@
+---
+"executor": patch
+---
+
+Fix toolkit MCP routes in the local daemon by reusing its owned SQLite database and sharing one scoped executor per toolkit instead of reacquiring storage and connector processes for every client.
--- a/apps/local/src/executor.ts
+++ b/apps/local/src/executor.ts
@@ -11,6 +11,7 @@ import {
   runSqliteDataMigrations,
   type AnyPlugin,
   type Executor,
+  type FumaDb,
 } from "@executor-js/sdk";
 import { collectTables } from "@executor-js/api/server";
 import { loadPluginsFromJsonc } from "@executor-js/config";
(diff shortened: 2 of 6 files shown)


RhysSullivan:
Closing in favor of #1551

PR #1626, opened 2026-08-17
Bind span header redaction to the hosted HTTP client

RhysSullivan:
Follow-up to #1623. Live verification showed post-deploy http.client spans still recording non-allowlisted headers: consumers capture the client at construction and run requests on fibers the host telemetry layers never reach, so the Context.Reference provided by layer did not make it to the tracer's getRef. The redaction now travels with the client itself (HttpClient.transformResponse wrapping each request effect in provideService), with the allowlist single-sourced in the sdk and the cloud telemetry layer reusing it for the server-side tracer. Includes a regression test that drives a request through the real hosted client and asserts redaction on the recorded span attributes.

### DIFF
--- a/.changeset/hosted-client-span-header-redaction.md
+++ b/.changeset/hosted-client-span-header-redaction.md
@@ -1,0 +1,5 @@
+---
+"@executor-js/sdk": patch
+---
+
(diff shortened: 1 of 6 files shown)


RhysSullivan:
Deploy preview failure was a latent main breakage: plugins/mcp compiles with lib DOM but not DOM.Iterable, so Headers.entries() and the iterable protocol were missing from its program. It never surfaced because CI served that package build from the turbo cache; this branch's sdk change invalidated the cache and forced a fresh compile. Fixed here by adding DOM.Iterable (the pairing hosts/cloudflare already uses).

PR #1648, opened 2026-08-18
Keep connection list health compact by default

baggiiiie:
## issue

see https://github.com/UsefulSoftwareCo/executor/issues/1647

## root cause

the root cause for this issue is, `connection.lastHealth` is copied directly to `connections.list` output, and its schema used the full `HealthCheckResult`. The existing `verbose` option controlled only the full OAuth scope string, not health diagnostics.

## fix

`connections.list` now outputs:

- Default or `verbose: false`: `status`, `identity`, and `checkedAt` only.
- `verbose: true`: the complete health result, including `httpStatus`, `detail`, and `responseSample`.

Connection detail, create, refresh, and HTTP API behavior remain unchanged. Unit and end-to-end tests cover both compact and verbose responses.

### DIFF
--- a/.changeset/lean-connection-health.md
+++ b/.changeset/lean-connection-health.md
@@ -1,0 +1,5 @@
+---
+"@executor-js/sdk": patch
+---
+
+Keep `connections.list` health output compact unless callers opt into diagnostics with `verbose: true`. Default list responses now retain only the health status, identity, and check timestamp; verbose responses continue to include HTTP status, diagnostic detail, and bounded upstream response samples.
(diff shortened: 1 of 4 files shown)


baggiiiie:
hmm the ci failures seem to be flaky tests? this PR only touches `core-tools.ts`, the failures shouldn't be caused by the diff

i think cloud2 has flaky test assertion, cloud5 could be some load-dependent bug

RhysSullivan:
good pr thanks

RhysSullivan:
yeah they're flakey tests, will fix

baggiiiie:
hmm hey Rhys @RhysSullivan one job in ci got timeout'ed at `bun install --frozen-lockfile`, issue with blacksmith runner?

https://github.com/UsefulSoftwareCo/executor/actions/runs/32340881974/job/96339586996?pr=1648

PR #1665, opened 2026-08-18
Cache JWKS across isolates and serve stale keys on refresh failure

RhysSullivan:
## Problem

`workos.session.local_verify` was missing its JWKS cache ~92% of the time. Every miss paid a fresh upstream fetch (p50 3s), and the tail crossed the 5s fetch timeout — so a slow key server surfaced as `AbortError` → `WorkOSError` → **500 on every authenticated API route**, and `McpJwtVerificationError` → **503 on `/mcp`**. Roughly 150 errors/day against 0–8/day before.

The cache implementation was not itself broken — `mcp.auth.jwt_verify` uses the same code and stays healthy (p50 0ms). The gap is structural: module-scope memory only helps while an isolate stays warm, and it gives a cold isolate nothing to fall back on when upstream is slow.

## Change

Two layers, both in `jwks-cache.ts`:

1. **Cross-isolate store.** Defaults to the Workers Cache API, injectable for tests, `null` to opt out. A cold isolate reads keys colo-locally instead of paying an upstream round trip.
2. **Stale-while-revalidate.** Past `ttlMs` a usable key set is served immediately and refreshed in the background; a *failed* refresh keeps serving the last good keys until `staleMaxMs` instead of failing the verify. Only a fully cold path (no memory, no store) blocks on the network.

A transient key-server blip should never read as an auth failure. That was the actual bug.

## Safety

Serving a stale key set is not the same as serving a stale token — tokens are still signature- and expiry-checked against it. Bounded by `staleMaxMs` (24h), and a token whose `kid` is absent still forces a real refresh before rejection, so a retired key can't be honoured indefinitely.

## Verification

- `format:check`, `lint` (0 errors), `typecheck` (44/44) clean.
- 108 tests pass in `src/auth` + `src/mcp`; 4 new cases cover cold-isolate reads, serve-stale-on-outage, the stale-window cutoff, and non-blocking revalidation.
- The 3 behavioural tests were confirmed to **fail against the previous implementation** (the blocking one hung the full 5s), so they pin the regression rather than the fix.
- The `caches.default` path was exercised under **real workerd** via Miniflare: a fresh module scope with the upstream *down* verified successfully with zero extra upstream calls, while the same scenario with `store: null` failed — isolating the store as the thing that fixed it.

## Follow-up (not in this PR)

I could not pin *why* the hit rate collapsed at the 2026-08-16 21:00 UTC deploy (`0b1739be6`). That deploy touched neither `workos.ts` nor `jwks-cache.ts`, and traffic volume was flat across the break. This change makes the path resilient regardless, but the trigger is still worth understanding.

The `executor.sh — tool operation regression` monitor is also firing on declined approval prompts (`ElicitationDeclinedError`); it wants the same `requires approval` filter the MCP-errors monitor already has.

### DIFF
--- a/apps/cloud/src/auth/jwks-cache.node.test.ts
+++ b/apps/cloud/src/auth/jwks-cache.node.test.ts
@@ -9,7 +9,7 @@ import {
   type KeyLike,
 } from "jose";
 
-import { createCachedRemoteJWKSet } from "./jwks-cache";
+import { createCachedRemoteJWKSet, type JwksStore, type StoredJwks } from "./jwks-cache";
 
 const issuer = "https://test-authkit.example.com";
 const audience = "client_test_fixture";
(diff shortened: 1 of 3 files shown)


RhysSullivan:
Rebased onto main, which picked up #1664's JWKS fetch counters. Reconciled rather than clobbered: the counters and `inspect()` shape are preserved.

One interaction worth flagging. #1664 annotates `jwks.fetched_during_verify` as `fetchCount` moving across the verify. Under stale-while-revalidate a *background* refresh also moves `fetchCount` while the caller waits on nothing — so that attribute would have read `true` for verifies that paid no latency, in the exact incident it was added to diagnose. Added `blockingFetchCount` (fetches a verify actually awaited) and `storeHitCount` (cold-isolate reads served by the store), and pointed the span at the blocking counter, plus a new `jwks.served_from_store`.

CI note: the `Test` job is red on this PR from `packages/plugins/openapi` (`spec-format-adapter.test.ts`, Google preset consent-scope). That is **pre-existing on main** — main has been red since d16abec9f (#1663), the test file is byte-identical to main, and this branch touches only three files under `apps/cloud/src/auth/`. Not introduced here, and not fixed here.

PR #1836, opened 2026-08-29
feat(plugin-onepassword): Support multiple named 1Password vault configurations per owner (#1791)

### LINKED ISSUE #1791: Support multiple named 1Password vault configurations per owner
## Problem

The 1Password credential provider currently supports one owner-scoped configuration containing one `vaultId`. Configuring another vault replaces the existing configuration, and secret resolution rejects references outside that selected vault.

This makes least-privilege setups awkward. A user can point Executor at a dedicated automation vault, but may also need selected credentials that remain canonical in a Private, shared, or work vault. Copying those items creates independent copies that do not remain synchronized, while selecting the broader vault exposes more items than Executor needs.

## Proposed behavior

Support multiple named 1Password vault configurations per owner, for example:

- `executor`
- `personal`
- `work`

Each configuration would retain its own authentication metadata, `vaultId`, and display name. Secret references or provider item identifiers should route through an explicitly selected named configuration so the existing vault boundary remains enforceable.

The Secrets UI should list configured 1Password vaults and provide an **Add 1Password vault** action instead of switching to only **Edit** and **Disconnect** after the first configuration.

## Acceptance criteria

- Add, edit, and remove multiple named 1Password configurations without overwriting another configuration.
- Browse provider entries within a specific configured vault.
- Resolve a secret only through its selected configuration and vault boundary.
- Continue redacting service-account tokens and never expose credential values through discovery APIs.
- Migrate an existing singleton 1Password configuration to a default named configuration without breaking existing references.

## Current implementation context

The current plugin uses a fixed provider key and stores a single blob under the fixed `config` key. Its configuration schema contains one `vaultId`, and `configuredVaultUri` refuses a fully qualified `op://` URI when its vault does not match that configured ID. This request likely needs named provider configurations rather than only an additional vault picker in the UI.

Adityakk9031:
Resolves #1791

### Problem
The 1Password credential provider previously supported only a single owner-scoped configuration containing one vaultId (or multi-vault list in a single connection). Configuring another vault or account replaced the existing configuration, preventing users from separating Personal vs Work accounts or targeting distinct service accounts without overwriting siblings.

### Solution
This PR introduces support for multiple named 1Password vault configurations per owner:

1. **Named Configuration Model & Persistence**:
   - `OnePasswordConfig` represents an individual named configuration containing `{ id, name, auth, vaultId, vaultName? }`.
   - `StoredOnePasswordConfig` persists owner configurations under `{ configs: OnePasswordConfig[] }`.
   - Backward-compatible normalization seamlessly upgrades existing legacy single-vault and multi-vault blobs assigning `id: "default"`.

2. **Ref & URI Resolution (`resolveConfiguredRef`)**:
   - Named configuration addressing: `op://<configId>/<itemId>`, `op://<configId>/<vaultId>/<itemId>`, `op://<configId>/<itemId>/<field>`.
   - Legacy vault addressing: `op://<vaultId>/<itemId>` and `op://<vaultName>/<itemId>`.
   - Strict vault isolation: references outside configured vaults fail with `outside-vaults`.
   - Ambiguity detection: bare item references matching multiple configured vaults report all matching configs and vaults.

3. **API & Static Tools Surface**:
   - Added `listConfigs` endpoint (`GET /onepassword/configs`) and static tool.
   - Updated `getConfig`, `removeConfig`, and `status` to accept an optional `id` parameter.
   - Full service-account token redaction across all external/UI-facing endpoints.

4. **React Settings UI (`OnePasswordSettings`)**:
   - Displays all configured 1Password vault connections in the settings card stack.
   - Independent **Edit** and **Disconnect** buttons per vault configuration.
   - Added "+ Add 1Password vault" action with single-vault selection dialog.

5. **Testing**:
   - Exhaustive test suite covering multi-config management, independent creation/updates/deletions, token redaction, legacy normalization, and URI resolution/ambiguity detection.

### DIFF
--- a/packages/plugins/onepassword/src/api/group.ts
+++ b/packages/plugins/onepassword/src/api/group.ts
@@ -52,7 +52,14 @@ const GetConfigResponse = Schema.NullOr(RedactedOnePasswordConfig);
 
 export const OnePasswordGroup = HttpApiGroup.make("onepassword")
   .add(
+    HttpApiEndpoint.get("listConfigs", "/onepassword/configs", {
+      success: Schema.Struct({ configs: Schema.Array(RedactedOnePasswordConfig) }),
+      error: [InternalError, OnePasswordError],
+    }),
+  )
+  .add(
     HttpApiEndpoint.get("getConfig", "/onepassword/config", {
+      query: Schema.Struct({ id: Schema.optional(Schema.String) }),
       success: GetConfigResponse,
       error: [InternalError, OnePasswordError],
     }),
@@ -66,12 +73,14 @@ export const OnePasswordGroup = HttpApiGroup.make("onepassword")
   )
   .add(
     HttpApiEndpoint.delete("removeConfig", "/onepassword/config", {
+      query: Schema.Struct({ id: Schema.optional(Schema.String) }),
       success: Schema.Void,
       error: [InternalError, OnePasswordError],
     }),
   )
   .add(
     HttpApiEndpoint.get("status", "/onepassword/status", {
+      query: Schema.Struct({ id: Schema.optional(Schema.String) }),
       success: ConnectionStatus,
       error: [InternalError, OnePasswordError],
     }),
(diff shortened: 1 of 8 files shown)


RhysSullivan:

PR #1852, opened 2026-08-29
fix(runtime-quickjs): offload QuickJS executions to worker threads in selfhost to prevent event loop blocking

### LINKED ISSUE #1616: Self-host: a CPU-bound `execute` blocks the whole server for up to 5 min (QuickJS runs on the main thread)
## The situation

Self-host runs `execute` in QuickJS on the server's main thread. `apps/host-selfhost/src/execution.ts` wires `makeQuickJsExecutor()` with defaults, `runtime-quickjs` evaluates with the sync WASM build, and the only bound is the in-sandbox interrupt handler at `timeoutMs` = 300 000. So a CPU-bound script owns the Node event loop until it finishes or hits the 5 min deadline. Nothing else on the instance is served in the meantime.

Measured on the official `v1.5.40` image, Linux amd64, one MCP client:

- `execute` with `while (Date.now() - t0 < 150000) {}` returned 200 after 150.3 s.
- during those 150 s, `/api/health` took 3.6 s and a concurrent `GET /api/oauth/clients` got no answer within 30 s. The web UI did not load either.
- same script at 75 s: same behaviour, returned 200 after 75.6 s.

I/O-bound scripts are fine — every `tools.*` call yields to the host, and since #1437 the deadline is suspended while a dispatch is in flight — so normal agent code never triggers this. It takes one careless loop (spin-poll on a job, an accidental O(n²) over a big payload) to stall a shared instance for up to five minutes for every user and every MCP client on it.

## What works today, and why it is not enough

- The 5 min interrupt eventually kills the script. That is a ceiling, not isolation.
- #1551 adds `EXECUTOR_SANDBOX_TIMEOUT_MS` to self-host. I'll set it to ~60 s the day it ships; it shortens the stall, it does not remove it.
- Cloud does not have the problem because it runs code in dynamic workers, isolated from the API process. `runtime-deno-subprocess` exists in the repo but no app depends on it.

## Ask

Run the self-host sandbox off the main thread: a `worker_threads` worker around the existing sync QuickJS build would be enough (or the asyncify variant with a periodic yield, or the Deno subprocess runtime). Keep the interrupt deadline as the backstop. Merging #1551 with a lower default than 5 min would help in the meantime.

Not asking for per-execution CPU accounting or resource quotas here — just that one `execute` cannot take the whole server with it.

Adityakk9031:
Fixes #1616

### Summary

Self-host executions in `@executor-js/runtime-quickjs` previously ran QuickJS synchronously on the Node.js/Bun server's main thread. When a CPU-bound execution was run (e.g. `while (true) {}`), it monopolized the main event loop until hitting `timeoutMs` (up to 5 minutes by default), starving health probes, OAuth sessions, web UI, and concurrent MCP requests.

This PR introduces optional worker thread execution to `@executor-js/runtime-quickjs` and enables it for self-hosted instances:
1. **Worker Thread Execution (`worker: true`)**: Added `worker?: boolean` option to `QuickJsExecutorOptions`. When enabled, `makeQuickJsExecutor` delegates execution to a dedicated `node:worker_threads.Worker` running `packages/kernel/runtime-quickjs/src/worker.ts`.
2. **Bidirectional IPC**: Tool invocations inside the QuickJS worker seamlessly bridge back to the parent thread's `toolInvoker` and yield results/errors back to the worker context.
3. **Cancellation & Preemption**: Interruption or timeout terminates the worker thread cleanly without hanging or stalling the host server's main event loop.
4. **Self-Host Enabled**: `SelfHostCodeExecutorProvider` in `apps/host-selfhost/src/execution.ts` sets `worker: true`.

### Verification

- Unit tests added to `packages/kernel/runtime-quickjs/src/index.test.ts` covering plain executions, bidirectional tool dispatches, tool error propagation, and CPU-bound loop preemption in worker threads.
- Vitest tests:
  - `packages/kernel/runtime-quickjs`: 32/32 tests passed (including all worker thread suites).
  - `apps/host-selfhost`: 137/137 tests across 24 test files passed.
- Gates:
  - `bun run typecheck`: 44/44 successful.
  - `bun run lint`: 0 warnings, 0 errors.
  - `bun run format:check`: All modified files formatted and verified.

### DIFF
--- a/apps/host-selfhost/src/execution.ts
+++ b/apps/host-selfhost/src/execution.ts
@@ -68,9 +68,10 @@ export const SelfHostCodeExecutorProvider: Layer.Layer<CodeExecutorProvider> = L
   CodeExecutorProvider,
   () => {
     const { sandboxTimeoutMs } = loadConfig();
-    return makeQuickJsExecutor(
-      sandboxTimeoutMs === undefined ? {} : { timeoutMs: sandboxTimeoutMs },
-    );
+    return makeQuickJsExecutor({
+      ...(sandboxTimeoutMs === undefined ? {} : { timeoutMs: sandboxTimeoutMs }),
+      worker: true,
+    });
   },
 );
 
--- a/packages/kernel/runtime-quickjs/package.json
+++ b/packages/kernel/runtime-quickjs/package.json
@@ -31,6 +31,7 @@
   },
   "scripts": {
     "build": "tsup && (tsc --declaration --emitDeclarationOnly --outDir dist --rootDir src || true)",
+    "pretest": "bun run build",
     "typecheck": "tsgo --noEmit",
     "test": "vitest run",
     "test:watch": "vitest",
--- a/packages/kernel/runtime-quickjs/src/index.test.ts
+++ b/packages/kernel/runtime-quickjs/src/index.test.ts
@@ -433,3 +433,72 @@ describe("quickjs executor", () => {
     }),
   );
 });
+
+describe("quickjs worker executor (worker_threads)", () => {
+  const workerExecutor = makeQuickJsExecutor({ worker: true, timeoutMs: 5_000 });
+
+  it.effect("runs plain code and returns results in a worker thread", () =>
+    Effect.gen(function* () {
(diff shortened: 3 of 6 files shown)


RhysSullivan:

PR #1854, opened 2026-08-29
fix(oauth): clamp client ownership on single-workspace hosts

### LINKED ISSUE #1850: Single-workspace hosts allow creating user-scoped OAuth clients that can never be used
On local/desktop (single org workspace, real LOCAL_SUBJECT), oauth.clients.create accepts owner: "user" clients, but every local connection is org-scoped, so those clients are categorically unusable. The guard then rejects them with cloud vocabulary ("A Workspace connection must use a Workspace app.") on a host that has no Workspace concept.

Fix is either refusing/clamping user client ownership on single-workspace hosts or rewording per host. Split from #1330 (part 3 residual).

Adityakk9031:
### Summary
On single-workspace hosts (local/desktop), all connections are org-scoped (`LOCAL_CONNECTION_OWNER = "org"`). Previously, `oauth.clients.create`, `oauth.clients.registerDynamic`, and `oauth.createClient` accepted `owner: "user"` clients, but those clients were categorically unusable because local connections were org-scoped and the cross-owner guard rejected them with cloud-specific wording ("A Workspace connection must use a Workspace app.").

This PR:
1. Adds `singleWorkspace?: boolean` to `ExecutorConfig` (defaulting to true when `subject === "local"` or explicitly set).
2. Clamps `owner: "user"` to `owner: "org"` in `createClient`, `registerDynamicClient`, and core tools (`oauth.clients.create`, `oauth.clients.registerDynamic`, `oauth.clients.createHandoff`, `oauth.clients.remove`) on single-workspace hosts.
3. Allows local connections on single-workspace hosts to use existing/legacy client rows without throwing the cross-owner error.
4. Adds test coverage in `packages/core/sdk/src/oauth-flow.test.ts` and `packages/core/sdk/src/executor.test.ts`.

Closes #1850.

### DIFF
--- a/apps/local/src/executor.ts
+++ b/apps/local/src/executor.ts
@@ -226,6 +226,7 @@ const createLocalExecutorLayer = (options: LocalExecutorOptions = {}) => {
         // (loopback localhost is correct + intended for the local CLI, but it
         // is wired explicitly here rather than relying on a hidden default).
         redirectUri: new URL("/api/oauth/callback", webBaseUrl).toString(),
+        singleWorkspace: true,
         // Built-in agent-facing tools (integrations / connections / policies).
         coreTools: {
           webBaseUrl,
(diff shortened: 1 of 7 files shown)


RhysSullivan:

PR #1855, opened 2026-08-29
feat(google): add Analytics Admin and Data presets

aryasaatvik:
## Summary

Add first-party Google Analytics Admin and Data presets so users can connect GA4 through the existing Google Discovery and standard-user OAuth flow. Both integrations request read-only Analytics access.

## Presets

| Preset | Discovery API | Representative operation | Health check |
| --- | --- | --- | --- |
| Google Analytics Admin | `analyticsadmin/v1beta` | `analyticsadmin.accountSummaries.list` | `accountSummaries.list` |
| Google Analytics Data | `analyticsdata/v1beta` | `analyticsdata.properties.runReport` | Omitted; the API has no stable cheap read |

Both presets:

- use `https://www.googleapis.com/auth/analytics.readonly` without write scopes;
- preserve their service-hosted Discovery URLs;
- participate in Google bundle OAuth and provider-service split mapping;
- appear in the integration picker.

## Validation

- `bun x vitest run src/providers/google/presets.test.ts src/providers/google/discovery.test.ts` — 27 passed
- `bun x vitest run src/planner.test.ts` — 30 passed
- `bun run typecheck --filter=@executor-js/plugin-openapi --filter=@executor-js/plugin-provider-service-split`
- changed-file `oxlint --deny-warnings`
- changed-file `oxfmt --check`
- `bun run lint:changelog-stubs`

### DIFF
--- a/.changeset/google-analytics-presets.md
+++ b/.changeset/google-analytics-presets.md
@@ -1,0 +1,5 @@
+---
+"@executor-js/plugin-openapi": patch
+---
+
+Add Google Analytics Data and Admin as first-party Google Discovery presets with read-only OAuth scopes and service-hosted Discovery URLs.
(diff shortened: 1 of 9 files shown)


aryasaatvik:
@greptile review

PR #1886, opened 2026-08-30
fix(cloud): require CSRF state in the WorkOS login callback

ra-co88:
## What

The WorkOS login callback now requires a valid, unconsumed CSRF `state` parameter on every request. Missing state, unknown state, or a replayed (already consumed) state each return 400 before any WorkOS API call is made.

## Why

Login CSRF: an attacker can craft a callback URL that logs the victim into the attacker's account. The `state` parameter ties the callback to a login flow the user actually initiated — without enforcing it unconditionally, the callback accepts forged authorization codes. The check now happens before any network call to WorkOS, so forged requests are rejected at zero cost.

## What changed

- The callback handler rejects requests with no `state` (400) before any WorkOS call.
- State values are single-use: a replayed state returns 400.
- A fresh state matching the browser's cookie proceeds to the session exchange (302).

## Breaking changes

Flows that initiate login server-side and construct the callback URL without a state parameter will now be rejected. If you drive login from a server, generate a random nonce, persist it server-side (or sign it), and pass it as `state` — the callback validates it against the cookie it sets.

## Test plan

Focused suite on the callback handler (`apps/cloud/src/auth/workos-callback-state.node.test.ts`):
- no state → 400, no WorkOS call issued
- replayed/consumed state → 400
- fresh state matching cookie → 302 + session

All green against current main.

### DIFF
--- a/.changeset/tidy-login-state.md
+++ b/.changeset/tidy-login-state.md
@@ -1,0 +1,17 @@
+---
+"@executor-js/cloud": patch
+---
+
+fix: make login CSRF state mandatory in the WorkOS callback
+
+The callback previously skipped its CSRF check whenever the redirect carried
+no `state` value ("some WorkOS-initiated redirects don't include one"). That
+bypass let an attacker complete their own OAuth round-trip and redirect a
+victim's browser through the callback with the attacker's `code` and no
+`state`, silently signing the victim into the attacker's account (login CSRF).
+
+The check is now unconditional: a callback without a state matching the
+`wos-login-state` cookie set on `/login` is rejected with 400. This is a
+breaking change for any client relying on the undocumented no-state entry
+path; server-initiated flows that cannot carry state must be redesigned with
+a signed nonce instead of re-adding the bypass.
--- a/apps/cloud/src/auth/handlers.ts
+++ b/apps/cloud/src/auth/handlers.ts
@@ -188,17 +188,17 @@ export const CloudAuthPublicHandlers = HttpApiBuilder.group(
           const workos = yield* WorkOSClient;
           const users = yield* UserStoreService;
           const cookieState = request.cookies[STATE_COOKIE] ?? null;
-          // CSRF check is only enforced when the redirect carries a state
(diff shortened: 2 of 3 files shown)


ra-co88 (commented):
Superseded by the corrected review below (this one was posted with shell-mangled formatting). See the 07:28 UTC review for the actual verdict.

ra-co88 (commented):
Verdict: approve — login-CSRF hardening, correctly implemented and tested. (Comment review: GitHub blocks formal self-approval on your own PR.)

Supersession check (upstream/main @ 2dc399e51): the callback still only enforces state when present — the `if (query.state !== undefined)` guard allows a no-state entry path, which is precisely the login-CSRF hole this PR closes. Not superseded; the fix is genuinely missing upstream.

The change is the right shape:

- **Unconditional state check** before any WorkOS network call — forged callbacks now die at 400 with zero WorkOS cost, and the code comment documents the attack (attacker completes their own round-trip, victim gets signed into the attacker's account).
- **Single-use contract preserved**: the state cookie is consumed on use; the replay test at workos-callback-state.node.test.ts:150 drives a second callback with the same state and no cookie and asserts 400.
- **Negative cases are real, not vacuous**: node-level (replay ⇒ 400), e2e-level (login-csrf.test.ts:69 replayed GET ⇒ 400 "Invalid login state"), plus browser-binding and provider-redirect capture coverage. The negative-control discipline holds.
- **Timing-safe comparison** on the state match.

One honest caveat for the maintainer: the branch's diff footprint vs upstream is large (34 files) because it carries an old merge-base; the actual payload is 4a0c27e8a (the handler hardening) plus its test commits. The extra files (session-build-semaphore.test.ts, org-api-keys-console.test.ts) arrive via the stale base, not this PR's intent.

CI: all shards green (cloud 1-16, local, selfhost). Merges clean against current upstream/main.

Minor note: the missing-state case compares against the empty string rather than branching separately; fine, though a maintainer may prefer an explicit two-branch form for readability. Not blocking.

PR #1895, opened 2026-08-30
fix(e2e): honor the 503 restart envelope in the cap-eviction helper

ra-co88:
## What

Adds a retry backstop to `openSession` in `cloud/mcp-session-cap-eviction.test.ts`: when the server answers `initialize` with the documented restart envelope (503, JSON-RPC -32001, `"MCP session is restarting, please retry"` — the contract `agent-handler` emits for a platform DO reset), the helper retries the same request after a short delay instead of failing the scenario on a retryable platform blip.

## Context

The original root-cause half of this PR (opening concurrency) was superseded by #1907, which landed sequential admission and has the shard green on main. What remains here is only the retry: the exact behavior a real streamable-http client follows for that response, kept so the test's failure mode stays aligned with client behavior if the envelope ever recurs.

## Verification

- `cloud/mcp-session-cap-eviction` passes locally on this branch (rebased onto current main).
- `oxlint` and `tsc --noEmit` clean on the touched package.

### DIFF
--- a/e2e/cloud/mcp-session-cap-eviction.test.ts
+++ b/e2e/cloud/mcp-session-cap-eviction.test.ts
@@ -73,21 +73,42 @@ const openSession = async (
   label: string,
   recordSession: (sessionId: string) => void,
 ): Promise<string> => {
-  const initialized = await postJson(mcpUrl, bearer, {
-    jsonrpc: "2.0" as const,
-    id: "initialize",
-    method: "initialize",
-    params: {
-      protocolVersion: PROTOCOL_VERSION,
-      capabilities: {},
-      clientInfo: { name: `executor-e2e-cap-eviction-${label}`, version: "0.0.1" },
-    },
-  });
-  const sessionId = initialized.headers.get("mcp-session-id");
-  if (!sessionId) {
+  // The platform can reset a session Durable Object while its initialize
+  // is in flight, and the server answers that with the documented restart
+  // envelope (503, -32001, "MCP session is restarting, please retry") — the
+  // same contract a streamable-http client follows: same request, after the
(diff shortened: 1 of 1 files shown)


ra-co88 (commented):
Verdict: approve — the option-2 shape, now verified end-to-end. (Posted as a comment: GitHub refuses a formal self-approval on your own PR, so this is the author-side sign-off for a maintainer to convert.)

Supersession check: #1907 already dropped the opening concurrency 8 to 1 on main, which is why the 8-to-2 hunk was dropped here — the trimmed branch keeps only the retry backstop, correct call.

The remaining change is the openSession retry on the documented restart envelope, and it is exact: 503 + code -32001 + message 'MCP session is restarting, please retry' matches packages/hosts/cloudflare/src/mcp/durable-object-errors.ts, and 503 = retry-same-id is the contract in apps/cloud/src/mcp/agent-handler.ts. The decoder is strict (Schema.Literal on all three fields) and covered by it.each negative cases (404 with restart body, 503 with wrong code, wrong message, malformed JSON) — the negative-control discipline holds.

With sequential admission (#1907) the envelope is a defensive backstop rather than a live-scenario fix, and that is fine: a one-session-at-a-time helper that still dies on a transient platform DO-reset would be a flake factory. 8 attempts × 2s (host-advertised Retry-After) is bounded and reasonable.

Note on the semaphore test change (session-build-semaphore.test.ts fake-timers hunk): deterministic-clock coverage for the queue-timeout path that the retry loop can now exercise — no runtime effect, keeps the backstop testable. All CI green across every shard including E2E cloud 13of16 (the shard this scenario lives in). Merges clean. Good to merge.

PR #2005, opened 2026-09-15
Read members from the directory and add email search

RhysSullivan:
Layer 4 of 5. Answers the customer request for email search on the org users page.

- Both hosts' admin user directories are now `adminUserDirectoryFromMembers` over the seam. Cloud's per-row WorkOS lookups are deleted.
- `?search=` on the admin users endpoints: substring match over email and name, then subjects for those ids with the caller's paging. SDK `listSubjects` gains an `externalIds` filter. Exact `?email=` still works.
- Cloud `listMembers` and seat counts read the mirror. An org whose row has no `backfilled_at` mark gets a one-time on-demand scan first, so counts are never partial.
- Search input on the admin users page.
- `OrgMember.email` is now nullable in the shared contract because a membership can be mirrored before its profile.

Deploy prerequisite (cloud): run `bun run --cwd apps/cloud db:backfill-workos-mirror:prod` before deploying and confirm its counts against WorkOS.


RhysSullivan:
Recreated with the correct base after the stack was reordered (reconciler now sits below readers).

PR #2007, opened 2026-09-15
Authorize org membership from the local mirror

RhysSullivan:
Layer 5 of 5. Cloud only. Takes WorkOS off the per-request path.

- `authorizeOrganization` reads `MemberDirectory.membership` instead of calling WorkOS. Active status required; role mapping unchanged. Admin gates on the account, admin, and org planes use the role the authorize step resolved. The membership ownership check on remove and role change is one point read. The org switcher and free-org limit read `membershipsOf`.
- The mirror is trusted only while it is ready: the backfill has completed and the reconciler has drained within the last ten minutes. Until then the check falls back to WorkOS exactly as before, so nobody is locked out and nobody revoked during an outage is let in. An org the mirror has never seen is resolved from WorkOS on demand for a caller WorkOS confirms as a member.
- Org deletion order is now mark deleted locally, cancel billing, delete in WorkOS, purge. A retry after a partial deletion is admitted from the mirror.
- Deploy gate: `deploy.yml` runs `scripts/ensure-workos-mirror-ready.ts` after migrations. It runs the backfill if needed, drains the events stream itself, and fails the deploy if the mirror is still not ready.
- Unused WorkOS wrapper methods removed.

Ops step (cloud): add `WORKOS_API_KEY` to the `production` GitHub environment so the deploy gate can run the backfill. Removal in the WorkOS dashboard now takes effect when the event lands, normally seconds and at most about a minute. Removal through Executor stays immediate.


RhysSullivan:
Recreated with the correct base after the stack was reordered (reconciler now sits below readers).

Issue #1979, opened 2026-09-12
Cloudflare host: partial tool catalogs, refresh-token reuse, and annotations dropped from describe.tool


daviesayo:
## Summary

Running `apps/host-cloudflare` in production (v1.6.8, Workers + D1 + R2), I hit three defects that share one shape: **state that is only correct on a single-process host silently degrades on Workers.** Two of them are visible to callers as "the integration lost its tools" and "the connection is degraded again"; the third is a smaller gap in what `describe.tool` reports.

All three are read from `main` at `f1d95f2` unless noted. I have a reproduction for the first one's arithmetic and live evidence for all three.

---

### What I observed

An MCP integration with **16 tools** silently became an integration with **7 tools**. Calling a missing one returned `tool_not_found` with the seven survivors offered as suggestions, so from the caller's side the server simply "did not have" `get_digest` any more. A `connections.refresh` restored all 16 immediately, twice.

The seven survivors were exactly the **first seven tools in the server's `tools/list` order**. Not a random subset, not an alphabetical one: a prefix.

### Why 7

`persistCatalog` replaces a connection's catalog as delete-then-insert (`packages/core/sdk/src/executor.ts`, `produceConnectionToolsUnshared`):

```ts
yield* persistCatalog(
  Effect.gen(function* () {
    yield* core.deleteMany("tool", { where });
    yield* core.deleteMany("definition", { where });
    yield* core.createMany("tool", toolRows);
    yield* core.createMany("definition", definitionRows);
    yield* stampSynced(existingRow);
  }),
);
```

`persistCatalog` wraps that in `transaction(...)`. On the Cloudflare host the wrapper is not a transaction:

```ts
// apps/host-cloudflare/src/db/d1.ts
// `interactiveTransactions: false` — D1 rejects interactive transactions, so
// the fuma adapter runs transaction callbacks directly (auto-commit per
// statement).
const { db: fumaDb, fuma } = createExecutorFumaDb(drizzleDb, {
  ...options,
  interactiveTransactions: false,
  maxBoundParameters: 100,
});
```

And `createMany` splits into batches sized by the bound-parameter budget (`packages/core/fumadb/src/adapters/drizzle/query.ts`):

```ts
const rowsPerStatement = Math.floor((budget - reservedParameters) / columnsPerRow);
```

[...]

Issue #1973, opened 2026-09-11
OAuth refresh: persist rotated refresh tokens (Granola MCP invalid_grant)


dsivaasi:
## Context

Workspace: asi-nz
Integration: granola (official MCP at https://mcp.granola.ai/mcp)
Connection: tools.granola.user.daniel
OAuth client: org DCR dcr-mcp-auth-granola-ai to mcp-auth.granola.ai

## Symptom

Connection lastHealth.status became expired. Live tool call returned oauth_reauth_required / invalid_grant (HTTP 400 from mcp-auth.granola.ai on refresh).

Access expiresAt was about 2026-09-02; refresh rejected when probed 2026-09-11 NZ.

## Why we think this is refresh-token rotation / persistence

Executor already attempts access-token refresh. Granola MCP OAuth commonly rotates the refresh token on each successful refresh. If the client does not persist the latest refresh_token from that response, the next refresh uses a stale refresh token and Granola returns invalid_grant, forcing a full browser re-auth.

## Ask

1. Confirm whether Executor persists rotated refresh tokens from refresh_token grant responses for MCP/DCR OAuth connections (especially WorkOS Vault-backed ones).
2. If not, please fix so the newest refresh_token from the token endpoint replaces the stored one atomically.
3. Optional: surface refresh failing / reauth soon before lastHealth expired so bots can oauth.start earlier.

## Not asking for

A headless forever-token. Granola MCP is browser OAuth per user. We only want refresh to keep working for the lifetime Granola actually grants.

Issue #1520, opened 2026-08-02
OAuth refresh dedup is scoped per execution stack, allowing cross-session token reuse


brendanlees:
## Summary

On self-hosted Executor v1.5.37, concurrent OAuth refreshes are deduplicated within one scoped Executor instance, but not across separate MCP execution stacks or sessions. With providers that rotate refresh tokens, two sessions can submit the same refresh token. The provider then rejects reuse and may revoke the entire token family.

This appears to be a cross-session gap in the concurrency fix from #367.

## Environment

- Official self-hosted v1.5.37 image
- Single Executor container and single SQLite database
- Encrypted-secrets credential provider
- Several toolkit MCP endpoints consumed by multiple long-lived and short-lived MCP clients
- One Executor replica, with no database restore during the reproduction

## Observed behavior

Four separate Todoist MCP connections were reauthorized. Each connection has its own DCR client and user partition; no OAuth client is shared between those connections.

For every connection:

1. The initial access token worked.
2. The first post-expiry refresh worked.
3. Executor persisted the returned rotated refresh token and advanced `expires_at`.
4. A later refresh cycle failed with:

```text
invalid_grant: refresh token reuse detected; tokens for this client/user/resource revoked
```

All four connections eventually reached the same state independently.

A separate self-hosted n8n MCP OAuth connection also refreshed successfully once, then later returned HTTP 401 from its refresh endpoint. A static-token MCP connection remains healthy. Notion OAuth currently continues to refresh, likely because Notion documents a one-token grace window for refresh rotation.

## Why this looks cross-session

The v1.5.37 implementation creates `refreshInFlight` inside the Executor instance:

- `packages/core/sdk/src/executor.ts`

A scoped Executor is built by `makeScopedExecutor`, and MCP server or session construction builds an execution stack containing a new scoped Executor:

- `packages/core/api/src/server/scoped-executor.ts`
- `packages/core/api/src/server/mcp-build.ts`

As a result, the per-connection `refreshInFlight` map is not shared by separate MCP execution stacks. Two stacks can both read the same stored refresh token and each believe they are the refresh winner.

This is particularly dangerous for rotating refresh tokens. Notion's own MCP documentation warns that refreshes must use a mutex or distributed lock across workers or replicas and that reuse can revoke the connection:

https://developers.notion.com/guides/mcp/build-mcp-client

## Reproduction outline

1. Run self-hosted Executor v1.5.37.
[...]

PR #2037, opened 2026-09-17
Fix the wrong Expired status and the refresh races behind it

spa5k:
## Summary

Connections showed a wrong **Expired** status — sometimes false, sometimes permanent — and the token refresh behind it was uncoordinated. This fixes the four defects that produced it, across `@executor-js/sdk` and the MCP, OpenAPI, and GraphQL plugins.

## What changes

**A lost refresh race no longer kills a connection.** The in-flight refresh gate is keyed on the database handle, so two instances can redeem one single-use refresh token. The loser received `invalid_grant` and wrote the permanent rejection record while the winner's valid rotated token sat in the store unused — every surface then answered `expired` without probing, and nothing but a re-authorization recovered it. The refusal now re-reads the stored refresh item: a value that changed during the request is a peer's rotation, so the call adopts the access token that peer persisted and records nothing. The record write is also skipped when `expires_at` moved forward during the grant.

**One rate-limited minute no longer ends a grant.** 408, 425, and 429 are retryable now, like a 5xx.

**The health probe refreshes before it answers `expired`.** A revoked token, an idle timeout, or a null `expires_at` used to show a working connection as dead until a tool call healed it. The probe now re-mints once and probes again for OAuth connections; a refused refresh keeps the probe's own verdict, so a dead grant still reads `expired`.

**The plugin is asked first, with or without a spec.** MCP's probe needs no spec, so its OAuth connections now get a real verdict. Only a plugin that answers `unknown` falls back to the credential-only verdict — computed from the values the probe already resolved, so nothing refreshes twice — and that fallback reports `expired` when a credential value resolves to nothing.

**The MCP liveness probe reuses the pooled connection.** It used to dial a second connection, which for a stdio server started a second child process that single-instance servers (Chrome DevTools MCP, Playwright MCP, `docker run -i`) refuse — reporting a live, serving connection as broken on every page mount. The pooled lease is bounded by the same 15s discovery deadline, and an interrupted probe still releases.

**A scope shortfall reads `degraded`, not `expired`.** Recognised from an RFC 6750 `WWW-Authenticate` challenge or a body marker; the remedy is a new consent, not a reconnect. The GraphQL probe no longer reads a transport failure's prose as a credential verdict (`connect EACCES: permission denied` is not one).

**A refresh without `expires_in` no longer erases the expiry.** RFC 6749 makes the field optional; writing null disabled proactive refresh permanently. The mint records the advertised lifetime and a refresh derives from it.

Also: the test OAuth server's `/mcp` endpoint now speaks JSON-RPC properly (request's own id, empty catalog for `tools/list`, silence for notifications). The old canned reply used a fixed id, so every completed handshake waited forever for `tools/list` — invisible until health checks started dialling.

## What does not change

The permanent rejection gate stays: a genuinely rejected grant still stops refresh traffic and reads `expired` until a reconnect. No new `HealthCheckReason` literal ships (the set is persisted and needs readers first). The credential-only fallback still reports `healthy` when nothing can probe — changing that turns a large class of connections grey and is a separate decision.

## Related issues

- **#1979, section 2** — the in-flight refresh gate does not dedup on Workers, so rotating refresh tokens get reused. This PR is that diagnosis: the same code comment, the same symptom (a connection flapping degraded and healing once one refresh succeeds alone). The fix here removes the *consequence* — the loser of a rotation race adopts the peer's token and records no rejection, so no `invalid_grant` surfaces and no verdict flips. The cross-isolate single-flight this issue proposes (the DO or database lease) is the remaining half and is not in this PR. Section 2's smaller item — `oauth_scope` keeping the requested set when the provider never echoes `scope` — is also not addressed here.
- **#1973** — Granola MCP `invalid_grant`, `lastHealth` expired, re-auth required. Executor does persist rotated refresh tokens (`persistRefreshedToken` writes the rotated token before the access token), so the reported failure matches the cross-isolate race above: the loser wrote the permanent rejection record while a valid rotated token sat in the store. This PR stops that from being written. An affected connection still needs one reconnect to clear an existing record, and the issue's third ask (surface re-auth earlier) is not addressed.
- **#1520** — the earlier, single-process instance of the same race; closed by sharing the gate across stacks. The gate's handle-keyed scope is what this PR works around.

## Verification

- [x] `bun run format:check`, `bun run lint`, `bun run typecheck` (45/45) — clean.
- [x] `bun run test` — 39/39 tasks (`TURBO_TEST_CONCURRENCY=4`).
- [x] e2e (selfhost): 17 scenario files, 46 tests — covers refused-refresh persistence, the reactive 401 retry, single-flight refresh across stacks, the freshness gate, the reconnect journey, the Google 403 carve-outs, GraphQL introspection health, and the MCP liveness path. Plus `local/oauth-await-longpoll` and `local/oauth-token-durability` on the local target.
- [ ] CI re-run needed for two flaky checks: `host-cloudflare`'s Access test (5s `unstable_dev` boot timeout; reproduced on `main` in a clean worktree under the same load; passes 3/3 in isolation here) and `mcp-session-cap-eviction` (also failed the tests-and-docs-only #2036, which has a fully green run of the same diff).

## Checklist

- [x] Changeset added (one, covering all four packages).
- [x] Tests added or updated for every behavior change.
- [x] No secrets, credentials, or private data in the diff.


spa5k:
@RhysSullivan Ready for review when you have a moment.

I hit this in day-to-day use: connections showed an Expired status at random, and local MCP servers (Chrome DevTools MCP) showed as disconnected while their tools still worked. The PR fixes the causes I could reproduce — the cross-isolate refresh race, the permanent rejection written from one rate-limited response, and the probe that answered without refreshing.

Two notes for review:
 - #1979 (section 2) and #1973 report the same failure class from production; the PR description links both.
 - Two CI checks failed with flake signatures (the host-cloudflare Access test times out on boot 
 - it reproduces on main in a clean worktree — and mcp-session-cap-eviction, which also failed a docs-only PR). Both need a re-run, which I can't do from my account.

spa5k:
1. It shows linear expired.
Screenshot 2026-09-18 at 12 45 05 AM
2. Clicking linear and coming back shows it logged in.
Screenshot 2026-09-18 at 12 45 14 AM


Supabase MCP as well.

spa5k:
Have seen same things for the chrome devtools mcp.

